PR #1, opened 2018-02-20
Simplify .travis.yml according to Travis Rust guide

xfix:
Nightly is prone to randomly breaking, I don't think it should break the build (although it currently failing is a bug).


samrayleung:
Thanks for your heads-up :)

PR #5, opened 2018-05-15
Fix failing tests

DomWilliams0:
I fixed a few issues that were preventing the tests from passing:

- GET requests with a body are rejected by Spotify with a 400 response - even `{}`
- If the argument to `parse_response_code` is an empty string, the function panics
- A typo in "compilation" causes deserialisation to fail

Cargo also keeps trying to update the version in `Cargo.lock` to 0.1.2, so I commited that too :)

Thanks a lot for this library, it has saved me a lot of time! :100:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -673,7 +673,7 @@ dependencies = [
 
 [[package]]
 name = "rspotify"
-version = "0.1.1"
+version = "0.1.2"
 dependencies = [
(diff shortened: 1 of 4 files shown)


DomWilliams0:
Welp, they still fail on Travis because it doesn't have a client ID/secret - if you add yours to its environment they should pass there too :^)

samrayleung:
Thanks for your head-up, and I am so glad you enjoy this crate :)

PR #8, opened 2018-06-12
replace error_chain crate with failure crate for error handling

maxjoehnk:
This change replaces the error_chain crate with the failure crate.
Primary reason for this change is that the old implementation isn't sync and therefore can't be sent between threads.


samrayleung:
This is a breaking change, it changes  the return value of function, such as:  from
```rust
pub fn artists(&self, artist_ids: Vec<String>) -> Result<FullArtists>
```
to 
```rust
pub fn artists(&self, artist_ids: Vec<String>) -> Result<FullArtists, failure::Error> 
```
I haven't taken thread-safe into consideration while I developed `rspotify`. I am not familiar with `failure` crate, why `rspotify` would be sync after replacing `error_chain` with `failure` :)

samrayleung:
I have upgraded `rspotify` to 0.2.2 to include this new feature, for more details, you need to check https://crates.io/crates/rspotify :)

PR #9, opened 2018-06-14
[feat] add debug and clone derives to spotify client and credentials

maxjoehnk:
This PR adds the `Debug` and `Clone` derives to `Spotify` and `SpotifyClientCredentials`.

Not sure whether we should add Serialize and Deserialize too.


samrayleung:
I think it will better to add Serialize and Deserialize, perhaps others will need this feature.

PR #10, opened 2018-07-12
Fix credentials error message appearing when using OAuth

oliveroneill:
The `build` method for  does not `SpotifyClientCredentials` does not consider `token_info` when printing out the "You need to set your Spotify API credentials" message. I've fixed this now. Let me know what you think.


samrayleung:
I think it is a mistaken bug, thanks for your heads-up.

PR #11, opened 2018-08-19
OAuth get_authorize_url fixes

oliveroneill:
I've made a few changes to the get_authorize_url function so that it better matches the spotipy implementation. I believe this was the intention here but it had a few lines in need of improvement.

The changes are:
- show_diaload changed to show_dialog. I believe this is a typo
- state parameter falls back to self.state if None
- show_dialog value is set before setting to "true" in parameters

I can split these changes up into multiple pull requests if necessary.


samrayleung:
merged, thanks for your heads-up.

PR #17, opened 2018-09-16
Remove Cargo.lock

jereksel:
According to docs (https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html) libraries shouldn't have Cargo.lock in repository


samrayleung:
I don't know this best practice rule before, thanks for your heads-up

PR #25, opened 2018-11-06
remove unneeded extern crate

uuttff8:
As title, I remove unneeded extern crate. I test code without it.


samrayleung:
thanks for your heads-up, man

PR #39, opened 2019-04-20
Fix de-serialization panics for null values

hrkfdn:
See also: https://github.com/samrayleung/rspotify/issues/37

I have also bumped the webbrowser dependency to allow rspotify to be built for *BSD systems.


samrayleung:
Thanks for your contribution, I don't realize `href` field and `id` field will be null in `Ablum` object, you do a great work :)

PR #40, opened 2019-04-28
FIx errors when email or birthdate are missing.

KoffeinFlummi:
The `current_user` endpoint was failing for me since both email and birthdate were missing.


samrayleung:
it's merged.

PR #41, opened 2019-05-29
Use log instead of print

braincow:
Use log crate macros instead of println! in multiple locations in code. This makes the console more readable since its not flooded with debug messages (unless so chosen with RUST_LOG=debug) all the time.


ramsayleung:
It's great, thanks for your heads-up, merged.

PR #42, opened 2019-05-29
get_token processing can be done in two parts

braincow:
This patch separates get_token functionality into two different functions while still maintaining legacy compatibility where user is queried on the command line for return URL address.

Separating the behavior allows handling request and verification of received token in code. For example: I am working on a small app that spawns a simple http server to localhost to wait for users browser to return back from Spotify website. Thus user never needs to enter the values themselves to console.

### DIFF
--- a/src/spotify/util.rs
+++ b/src/spotify/util.rs
@@ -63,27 +63,33 @@ pub fn convert_str_to_map(query_str: &mut str) -> HashMap<&str, &str> {
     map
 }
 
+pub fn request_token(spotify_oauth: &mut SpotifyOAuth) {
+    let state = generate_random_string(16);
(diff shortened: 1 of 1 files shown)


braincow:
This is the app where I am testing the feasibility of this type of approach: https://github.com/braincow/scotty

ramsayleung:
Great, it looks clearer after refactoring `get_token` into two functions. And it seems there is a conflict between master and your branch, could resolve it first ?

braincow:
@ramsayleung Conflict resolved now. (...and it appears that I caused it with my previous logging related pull request. "My bad" :D)

ramsayleung:
It's great, merged ~~

PR #44, opened 2019-07-21
Make PrivateUser country optional

Voultapher:
Using the crate got a deserialize error calling `current_user`. The docs say this parameter is optional, so the struct should reflect that. Note this is a sever breaking change and should bump the version to 0.6.

### DIFF
--- a/src/spotify/model/user.rs
+++ b/src/spotify/model/user.rs
@@ -24,7 +24,7 @@ pub struct PublicUser {
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub struct PrivateUser {
     pub birthdate: Option<NaiveDate>,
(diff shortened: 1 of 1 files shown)


Voultapher:
No idea about  the CI failures, getting the same locally on master ¯\_(ツ)_/¯

ramsayleung:
> No idea about the CI failures, getting the same locally on master `¯\_(ツ)_/¯`

It seems Travis doesn't allow developer to access encrypted environment variable by pulling request, as the lacking of nessary variable, it makes sense about the CI failures.

ramsayleung:
Thanks for your reminding, I have read the Spotify docs again, and it seems the `country` field is only available when the current user has granted access to the user-read-private scope

PR #46, opened 2019-08-08
Add a missing reference to current_user_recently_played

togatoga:
Hi @ramsayleung 

Thank you for your great project. I really like this library.  I found a missing reference for current_user_recently_played.
`Spotify` will be consumed after I call this function.


ramsayleung:
It seems this is my mistake, Thanks for your heads-up, this PR is merged :)

PR #48, opened 2019-08-20
Remove an unnecessary mut from current_user_saved_tracks_delete

togatoga:
I found an unnecessary `mut` from current_user_saved_tracks_delete.  It is better to change functions like `current_user_saved_tracks_delete` and `current_user_saved_albums_delete` to have similar arguments.
```Rust
pub fn current_user_saved_albums_delete(&self, album_ids: &[String]) -> Result<(), failure::Error>
{
//..
}
pub fn current_user_saved_tracks_delete(&self, track_ids: &[String]) -> Result<(), failure::Error>
{
///..
}
```

I removed it and also modified example and test code.


ramsayleung:
It's Merged :)

Issue #51, opened 2019-09-01
Start playback at `position_ms`


Rigellute:
Firstly, thank you for creating this amazing crate, it's made working with the spotify API so fun. 

I have this usecase 
> As a user, I should be able to toggle play/pause and resume playback at the previous position.

Looks like the spotify api let's you pass in `position_ms` to the `play` endpoint https://developer.spotify.com/documentation/web-api/reference/player/start-a-users-playback/

Which I think is what I need?

#### Potential solution
Pass `position_ms` as a fourth paramter to the `start_playback` function.

Or perhaps I am missing something? Is there an existing way of resuming playback?

ramsayleung:
No, you didn't do anything wrong or miss something, it's just my mistake. It seems when I develop this crate, Spotify doesn't provide such an argument, I will fix it later.

ramsayleung:
And just feels free to make a PR to fix this issue, it's easy :)

Rigellute:
Cool thanks. Have raised a PR adding this 👍

PR #52, opened 2019-09-03
Add `position_ms` to `start_playback`

Rigellute:
This would let the user resume playback after pausing. 

Link to issue https://github.com/ramsayleung/rspotify/issues/51


ramsayleung:
It's cool, thanks for your contribution, merged :)

PR #57, opened 2019-10-09
add missing device types

epwalsh:
This just add a missing device type `TV`

### DIFF
--- a/src/spotify/senum.rs
+++ b/src/spotify/senum.rs
@@ -1214,6 +1214,9 @@ impl fmt::Debug for SearchType {
 
 
 /// Device Type: computer, smartphone, speaker, TV, etc.
+/// See the [Spotify developer
+/// docs](https://developer.spotify.com/documentation/web-api/reference/player/get-a-users-available-devices/#device-types)
+/// for more information, or in case we are missing a device type here.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 pub enum DeviceType {
     Computer,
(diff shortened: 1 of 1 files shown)


ramsayleung:
From the Doc about [Device types](https://developer.spotify.com/documentation/web-api/reference/player/get-a-users-available-devices/#device-types), it seems more devices type which I simply ignore should be added to the `DeviceType` enum

src/spotify/senum.rs
```diff
@@ -1219,5 +1219,6 @@ pub enum DeviceType {
     Computer,
     Smartphone,
     Speaker,
+    TV,
```
ramsayleung: Perhaps You should update the related Doc about Device type in case of confusing other developers.

epwalsh: Good point. I just added the remaining device types and put a link in the docs.

epwalsh:
@ramsayleung could we bump the version and publish to crates with this patch? Need this patch to resolve a bug on spotify-tui: https://github.com/Rigellute/spotify-tui/issues/39

ramsayleung:
> @ramsayleung could we bump the version and publish to crates with this patch? Need this patch to resolve a bug on spotify-tui: [Rigellute/spotify-tui#39](https://github.com/Rigellute/spotify-tui/issues/39)

Of course we could, I will bump the version to 0.7.0 now :)

ramsayleung:
Well, there we are, rspotify is [0.7.0](https://crates.io/crates/rspotify)

src/spotify/senum.rs
ramsayleung: Cool

epwalsh:
Thanks @ramsayleung!

PR #61, opened 2019-10-24
Fix tuneable attribute passing for recommendations

sral:
Currently only Values in the payload that contain String will be considered/added
because calling as_str() on Values containing other types return None, I think.

See: https://docs.serde.rs/serde_json/value/enum.Value.html#method.as_str

### DIFF
--- a/src/spotify/client.rs
+++ b/src/spotify/client.rs
@@ -1373,17 +1373,15 @@ impl Spotify {
                           "tempo",
                           "time_signature",
                           "valence"];
-        let prefixs = vec!["min_", "max_", "target_"];
-        for (attribute, prefix) in iproduct!(attributes, prefixs) {
+        let prefixes = vec!["min_", "max_", "target_"];
+        for (attribute, prefix) in iproduct!(attributes, prefixes) {
             let param = prefix.to_owned() + attribute;
             if let Some(value) = payload.get(&param) {
-                if let Some(value_str) = value.as_str() {
-                    params.insert(param, value_str.to_owned());
-                }
+                params.insert(param, value.to_string());
             }
         }
         // for attribute in attributes {
-        //     for prefix in prefixs {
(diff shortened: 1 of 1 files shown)


src/spotify/client.rs
```diff
@@ -1373,17 +1373,15 @@ impl Spotify {
...
-        let prefixs = vec!["min_", "max_", "target_"];
-        for (attribute, prefix) in iproduct!(attributes, prefixs) {
+        let prefixes = vec!["min_", "max_", "target_"];
+        for (attribute, prefix) in iproduct!(attributes, prefixes) {
             let param = prefix.to_owned() + attribute;
             if let Some(value) = payload.get(&param) {
-                if let Some(value_str) = value.as_str() {
-                    params.insert(param, value_str.to_owned());
-                }
+                params.insert(param, value.to_string());
```
ramsayleung: I am a little confused, do you mean if rspotify use `as_str()` instead of `to_string()`, only String type value could be inserted into `params`, other types could be ignored? So this patch is going to fix this case? I hope I get your point.

sral: The problem is that the rspotiy API accepts any valid JSON value but
unless you explicitly pass String the value/payload is silently
dropped and never sent to the Spotify Web API. For example if you do this:

https://github.com/ramsayleung/rspotify/blob/master/examples/recommendations.rs#L40-L41

The example will return recommendations but `min_energy` and `min_popularity` is
silently ignored/never sent to the API leading to recommendations that don't respect
the attributes passed in. This leads to "bugs" that are very hard to spot/understand, I had to read the code to understand why I was getting bad recommendations. 

Hope this helps explain what I'm trying to fix :)

ramsayleung:
Yes, it makes everything clear. Strictly speaking, it is a bug, thanks for your explanation :-)

PR #62, opened 2019-11-06
Provide more informational error strings for API

sputnick1124:
Downstream consumers of the API will now be able to get a bit more information about why their request has failed by simply looking at the string.

If this is not an appropriate place to put such information, feel free to reject the request.


ramsayleung:
It's good to has more detailed messages about why their request has failed

PR #66, opened 2019-12-01
Add support for checking that current user saved/followed given a…

echoSayonara:
Checking saved/followed tracks and playlists have been implemented, however, it is not implemented for albums and artists. I added these features in this PR.


ramsayleung:
Cool, it's a good job to implement some missing functions, and it would be better if could you add a related example to demonstrate how to use it.

echoSayonara:
I added examples, please check it.

PR #69, opened 2020-01-17
Add proxy setting

kunerd:


kunerd:
The current solution won't work well, because I missed that you can add multiple proxies to `reqwest`s `Client`. But it shows each place where the proxy configuration is currently needed. I think we need a central place from which every component that needs a `Client` instance can get one with all basic settings.

kunerd:
According to the [`reqwest` docs](https://docs.rs/reqwest/0.10.1/reqwest/#proxies), system Proxies are recognized by default in newer versions. So simple bumping up the request version should do the trick :).

ramsayleung:
I am just gonna update `rewqest` to latest version, but it seems need some works to fix up the problems caused by `async/await` support. Even though I have created an issue to announce that I am gonna bump up the request version, but I am too busy to do this. Hence, if you have time to do this job, feel free to send a PR :)

ramsayleung:
I have updated `reqwest` to the latest version, now rspotify supports system proxy by default(Notes that `reqwest` system proxy doesn't support socks proxy for now, check [this issue](https://github.com/seanmonstar/reqwest/issues/790) for more details). For more details, please check this PR https://github.com/ramsayleung/rspotify/pull/70.

Issue #79, opened 2020-02-24
The workload of adding support for `async/await`.


ramsayleung:
I have refactored the core code of `Rspotify` to support `async/await`, and refactored the project structure to shorten import path either. Now the foundation is done, it's time to build the roof, and it'll be great if you have time to help me do this: 
The following endpoint functions need to be `asynced/awaited` before Rspotify can take advantage of `async/await`, check this [branch: ramsay/support-for-async-await](https://github.com/ramsayleung/rspotify/tree/ramsay/support-for-async-await) for more details:

- [x] artist
- [x] artists
- [x] artist_albums
- [x] artists
- [x] artist_albums
- [x] artist_top_tracks
- [x] artist_related_artists
- [x] album
- [x] albums
- [x] search_album
- [x] search_artist
- [x] search_track
- [x] search_playlist
- [x] album_track
- [x] user
- [x] playlist
- [x] current_user_playlists
- [x] user_playlists
- [x] user_playlist
- [x] user_playlist_tracks
- [x] user_playlist_create
- [x] user_playlist_change_detail
- [x] user_playlist_unfollow
- [x] user_playlist_add_tracks
- [x] user_playlist_replace_tracks
- [x] user_playlist_recorder_tracks
- [x] user_playlist_remove_all_occurrences_of_tracks
- [x] user_playlist_remove_specific_occurrenes_of_tracks
- [x] user_playlist_follow_playlist
- [x] user_playlist_check_follow
- [x] me
- [x] current_user
- [x] current_user_playing_track
- [x] current_user_saved_albums
- [x] current_user_saved_tracks
- [x] current_user_followed_artists
- [x] current_user_saved_tracks_delete
- [x] current_user_saved_tracks_contains
- [x] current_user_saved_tracks_add
- [x] current_user_top_artists
- [x] current_user_top_tracks
- [x] current_user_recently_played
- [x] current_user_saved_albums_add
- [x] current_user_saved_albums_delete
- [x] current_user_saved_albums_contains
- [x] user_follow_artists
- [x] user_unfollow_artists
- [x] user_artist_check_follow
- [x] user_follow_users
- [x] user_unfollow_users
- [x] featured_playlists
- [x] new_releases
- [x] categories
- [x] recommendations
- [x] audio_features
- [x] audios_features
- [x] audio_analysis
- [x] device
- [x] current_playback
- [x] current_playing
- [x] transfer_playback
- [x] start_playback
- [x] pause_playback
- [x] next_track
[...]

Issue #78, opened 2020-02-23
Add support for `async/await`


echoSayonara:
I tried to add support for `async/await` in my [branch](https://github.com/echoSayonara/rspotify/tree/support-for-async-await). I added an example (`examples/async-artist-top-tracks`) for it and went well.

If you have no future or ongoing plans to add support for `async/await`, please try my branch and leave me feedback.

ramsayleung:
As for ongoing plan to support `async/await`, I have talked with the author of `spotify-tui` on this [PR](https://github.com/ramsayleung/rspotify/pull/72), and everything is working in process now. Because my full time job is so damn busy, I have to take some extra time to support `async/await`, so please be patient :)

echoSayonara:
Sorry, i did not mean to rush you.
Understood. I'm looking forward to it. I would like to be helpful for it if I can.

ramsayleung:
I think your work/branch is inspired, there are a lot of things I could learn from your work. I just push my WIP work to [remote origin](https://github.com/ramsayleung/rspotify/tree/ramsay/support-for-async-await). Instead of creating a `nonblocking` package, I prefer to create a `blocking` package, and make `nonblocking` default.

ramsayleung:
> I would like to be helpful for it if I can.

If you have time, you could help me to convert those endpoint functions to `async/await` version, I just update the internal functions to `async`, there are still a lot of endpoint functions that needs to update.
https://github.com/ramsayleung/rspotify/blob/2ed470567990e0c2241e88db9c79afc5566f9b01/src/spotify/client.rs#L172
 https://github.com/ramsayleung/rspotify/blob/2ed470567990e0c2241e88db9c79afc5566f9b01/src/spotify/client.rs#L207

echoSayonara:
will do.

ramsayleung:
I just submit all the work `Rspotify` needs to do to support `async/await`, if you have time to join me, feel free to check it [The workload of adding support for `async/await`](https://github.com/ramsayleung/rspotify/issues/79)

echoSayonara:
I started to modify endpoint functions to support `async/await ` in `client.rs`. I will send PR when i finished to implement them. Please wait.

ramsayleung:
Perhaps you could send a new PR once you have finished to implement a endpoint function, or you could mark which endpoint function you are going to modify? just in case of any conflicting situation that we are both modifying the same endpoint function.

echoSayonara:
[...]

PR #80, opened 2020-02-25
Implement async/await on all methods

Rigellute:
This is going to be a large pull request (we could merge this into a side branch?).

TODO
- [x] Implement async/await on all methods
- [x] Update examples
- [x] Fix tests

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,8 @@ log = "0.4"
 percent-encoding = "1.0.1"
 rand = "0.6.5"
 random = "0.12.2"
-reqwest ={ version = "0.10.1", features=["json","blocking","socks"]}
+reqwest = { version = "0.10", features = ["json", "socks"] }
+tokio = { version = "0.2", features = ["full"] }
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
--- a/examples/album.rs
+++ b/examples/album.rs
@@ -3,7 +3,8 @@ extern crate rspotify;
 use rspotify::spotify::client::Spotify;
 use rspotify::spotify::oauth2::SpotifyClientCredentials;
 
-fn main() {
+#[tokio::main]
+async fn main() {
     // Set client_id and client_secret in .env file or
     // export CLIENT_ID="your client_id"
     // export CLIENT_SECRET="secret"
@@ -18,6 +19,6 @@ fn main() {
         .client_credentials_manager(client_credential)
         .build();
     let birdy_uri = "spotify:album:0sNOF9WDwhWunNAHPD3Baj";
-    let albums = spotify.album(birdy_uri);
+    let albums = spotify.album(birdy_uri).await;
     println!("{:?}", albums);
 }
--- a/examples/album_tracks.rs
+++ b/examples/album_tracks.rs
@@ -3,7 +3,8 @@ extern crate rspotify;
 use rspotify::spotify::client::Spotify;
 use rspotify::spotify::oauth2::SpotifyClientCredentials;
 
-fn main() {
+#[tokio::main]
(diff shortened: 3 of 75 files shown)


ramsayleung:
ooh,  it seems we are doing the same job to add support for async/await(something is conflicting, but something is not). Would you like merge your branch into my WIP [branch: support-for-async-await](https://github.com/ramsayleung/rspotify/tree/ramsay/support-for-async-await). By the way, I think it will be better you will need to check these two issues first:
https://github.com/ramsayleung/rspotify/issues/79
https://github.com/ramsayleung/rspotify/issues/78

Rigellute:
Hehe whoops! I should have checked the issues before I started (only checked pull requests).

I actually learnt a lot about async/await in Rust by doing this so still happy to have done it.

I've got my branch to the point where `cargo check` passes, the tests also all compile now (including all examples) but the async tests are a little flaky (guessing there is a timeout or something). Different tests fail on each run.

And sure, if my work is mergable we can merge into your branch!

Rigellute:
The most interesting part (for me at least) was updating the `internal_call` method to leverage generic json parsing from the response - this let's us not need to the `convert_result` method.
https://github.com/ramsayleung/rspotify/pull/80/files#diff-36ae331d0439539ddec0f366a67f90cbR136

Rigellute:
Also, I nearly got carpal tunnel during this refactor 😄 .

Luckily vim macros came to the rescue.

Only running tests are failing now, which is great!

Any ideas on how to fix this?

```
dispatch dropped without returning error
```

This looks like a timeout

ramsayleung:
> The most interesting part (for me at least) was updating the internal_call method to leverage generic json parsing from the response

Wow, thanks for your head-up,  it's amazing to take advantage of generic from the reqwest response. 

> I actually learnt a lot about async/await in Rust by doing this so.

So do I

> And sure, if my work is mergable we can merge into your branch!

Definitely Yes, you could merge into my branch, you have done a huge work for me. Would you like to resend a PR to my branch: `support-for-async-await`? Because I am not only updating the old endpoint functions to `async/await`, but also keep the old blocking endpoint functions with a new additional feature `blocking`, then other developers could choose API by their flavor.

ramsayleung:
> Luckily vim macros came to the rescue.

Perhaps Emacs could save you either(just a joke).

> Any ideas on how to fix this? dispatch dropped without returning error

which test is failed?

Rigellute:
Awesome @ramsayleung! Will send this PR to your branch! 
Just noticed that the code in the docs is also failing, will fix now (this is such an amazing feature of Rust)

ramsayleung:
Notes that I think the next release of `Rspotify` will be a huge break change because of the support for `async/await`, which definitely breaks backward compatibility. So I decide to make an other break change into the next release, just to shorten the import path:
before:
```
use rspotify::spotify::client::Spotify;
use rspotify::spotify::oauth2::SpotifyClientCredentials;
```
 
after:
```
use rspotify::client::Spotify;
use rspotify::oauth2::SpotifyClientCredentials;
```

the `spotify` mod is unnecessary, so I just remove it. Maybe it will confuse you when you make a new PR to merge into my branch.

Rigellute:
All the tests are passing! Looks like async/await is good to go!

Thanks for the heads up regarding the import paths.

Your plan to keep the blocking endpoints as a feature is a great idea.

ramsayleung:
Why do you change the base branch back to `master` branch?

Rigellute:
I am leaving this branch as is for now for my own reference and to preserve things in case of a bad merge/rebase. 

I forked this branch and applying it to your branch.

Rigellute:
Closing this in favour of #81

PR #84, opened 2020-02-25
Revert result conversion but make use of async/await

Rigellute:
The pre-existing `convert_result` worked better for handling requests that have no response, so I've reverted to that implementation but added async/await again.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -197,7 +197,8 @@ impl Spotify {
     where
         T: for<'de> serde::Deserialize<'de>,
     {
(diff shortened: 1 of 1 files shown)


Rigellute:
Have tested this in `spotify-tui` and it's working well!

ramsayleung:
> The pre-existing convert_result worked better for handling requests that have no response,

what happen when leverage generic json to handle no response case ?

Rigellute:
You're right @ramsayleung, I could have passed down `None` type for no response requests but on reflection I think your previous architecture was clearer, more explicit and more flexible.

PR #85, opened 2020-03-16
Add to queue endpoint

BKitor:
Adding an API endpoint for adding songs to user's queue

### DIFF
--- a/examples/add_to_queue.rs
+++ b/examples/add_to_queue.rs
@@ -11,16 +11,20 @@ async fn main() {
         .build();
 
     match get_token(&mut oauth).await {
+        // Some auth stuff
         Some(token_info) => {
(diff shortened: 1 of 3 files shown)


ramsayleung:
Wow, it's great to see you guys add the missing endpoint, but it seems a test is failed :(

BKitor:
The tests seem pretty inconsistent, I think there's an error with the web API timing out due to too many requests. 
When running tests locally, there's usually only 1 test that fails, but it's never the same one, and sometimes they all pass. 
We might want to tweak the test script to make it more consistent.

ramsayleung:
>  We might want to tweak the test script to make it more consistent.
Good point, it's not that easy to build tests dependent on 3rd services(such as Spotify), I still have no idea to do it.

PR #88, opened 2020-03-29
Reqwest blocking feature

cosmoduff:
This fixes issue 87 and allows rspotify to build with the blocking feature enabled. I haven't done many pull requests so let me know if you need anything else.


ramsayleung:
thanks for your heads-up, you do a great job to fix my mistake :)

Issue #63, opened 2019-11-09
Get Player Error Reasons


sputnick1124:
Over in [`spotify-tui`](https://github.com/Rigellute/spotify-tui), I am getting API errors fairly frequently. Much of the time, the error code is 403 (I am able to see that now since https://github.com/ramsayleung/rspotify/pull/62), which likely means that I am getting a [player error](https://developer.spotify.com/documentation/web-api/reference/object-model/#player-error-object), but I get no more information than that.

I'm not great at web stuff; I don't understand what is happening to this object in the request response. I figured I would be able to define a struct to hold the player error:
```rust
struct PlayerError {
    status: u16,
    message: String, 
    reason: String,
}
```
and then (deserialize the json)[https://docs.rs/reqwest/0.10.0-alpha.1/reqwest/struct.Response.html#method.json] from the response directly into it, but `reqwest` seems to be hiding the json from me somehow. What's happening to it? How can I get access to the error reason? This would help immensely in providing more information about failure to both developers and end-users.

I'm happy to implement any changes that need to happen, but I don't have much experience handling web requests, and none of that experience is with Rust/`reqwest`, so any other devs here that could shed some light would be appreciated.

sputnick1124:
For reference, this is what I see when interacting with the endpoint through the (API 
```json
{
  "error": {
    "status": 404,
    "message": "Player command failed: No active device found",
    "reason": "NO_ACTIVE_DEVICE"
  }
}
```

The `Response` object from `reqwest`, however, does not seem to provide any access to that json object. It would seem I am potentially missing something obvious, but I don't know what.
```
Response { 
    url: "https://api.spotify.com/v1/me/player/play?device_id=4a205e1455078ac27724255dd891265738f17dfe", 
    status: 404, 
    headers: {
        "content-type": "application/json; charset=utf-8",
        "cache-control": "private, max-age=0",
        "access-control-allow-origin": "*",
        "access-control-allow-headers":"Accept, App-Platform, Authorization, Content-Type, Origin, Retry-After, Spotify-App-Version",
        "access-control-allow-methods": "GET, POST, OPTIONS, PUT, DELETE, PATCH",
        "access-control-allow-credentials": "true",
        "access-control-max-age": "604800",
        "transfer-encoding": "chunked",
        "date": "Sat, 09 Nov 2019 13:41:32 GMT",
        "via": "1.1 google",
        "alt-svc": "clear"
    }
}
```

ramsayleung:
[...]

PR #91, opened 2020-04-23
Handle regular and player error objects

sputnick1124:
Addresses #63.

There is a message given alongside most API errors, but it has been
thrown away until now. This commit presents the full JSON object to the
end consumer of the error (see [error object](https://developer.spotify.com/documentation/web-api/reference/object-model/#error-object), [player error object](https://developer.spotify.com/documentation/web-api/reference/object-model/#player-error-object))

The `From<&Response>` impl was removed and replaced with a non-trait `from_response` method so that the async `Response::json` method could be called.

### DIFF
--- a/src/blocking/client.rs
+++ b/src/blocking/client.rs
@@ -40,10 +40,21 @@ lazy_static! {
     pub static ref CLIENT: Client = Client::new();
 }
 /// Describes API errors
-#[derive(Debug)]
+#[derive(Debug, Deserialize)]
 pub enum ApiError {
     Unauthorized,
     RateLimited(Option<usize>),
+    #[serde(alias = "error")]
+    RegularError {
+        status: u16,
+        message: String,
+    },
+    #[serde(alias = "error")]
+    PlayerError {
+        status: u16,
+        message: String,
+        reason: String,
+    },
     Other(u16),
 }
 impl failure::Fail for ApiError {}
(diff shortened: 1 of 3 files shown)


sputnick1124:
Just realized I neglected to implement the blocking version... That will follow shortly. Also, I'm not sure about the failing test(s). I get random test failures when running locally on the master branch, so I don't think it's necessarily anything I've done here, but please correct me if I'm wrong.

ramsayleung:
>  That will follow shortly.
Looking forward to it.
>  Also, I'm not sure about the failing test(s)
I think the failing tests has nothing related to your PR, the tests are unstable for which they are depending on Spotify service, which will fail from time to time.

sputnick1124:
I'm not stoked about changing the `From<Response> for ApiError` to a move instead of by reference, but it was the easiest way to go about it. I don't actually see any problems with this at the moment, but if there is something I missed, let me know and we can work around it.

sputnick1124:
I'm getting lint errors in files I didn't touch. To keep this PR focused, I'm not planning on worrying about those if that's alright.

ramsayleung:
The tests fail because of the change of Spotify API, I am gonna fix it, and you could merge the upstream branch before you merge into master. See this PR for more details https://github.com/ramsayleung/rspotify/pull/94

sputnick1124:
Could you clarify what you need before this gets merged? Happy to help in any way, but it seems that this should be good to go as is.

ramsayleung:
This PR is merged, actually the failed tests has nothing to do with this PR, if you are curious about the failed tests, you could track on this PR: https://github.com/ramsayleung/rspotify/pull/94

PR #94, opened 2020-04-28
Fix deserializing Audio Analysis response

### LINKED ISSUE #93: Timestamp as string rather than u64
Please see the issue here, it appears that Spotify has switched their timestamp from u64 to a string.

https://github.com/Rigellute/spotify-tui/issues/416

Rigellute:
It seems the spotify API has changed their responses from the audio_analysis endpoint.

A number of fields seem to be missing and/or have different types now - some numbers are now contained in strings e.g. `timestamp`.

I've gone the heavy handed route and made all the fields optional. I started trying to only make certain fields optional but this was hit and miss as I tried other tracks.

Perhaps there is a better solution?

The errors can be reproduced by running
```
cargo run --example audio_analysis
```

This should close #93


ramsayleung:
Why does almost all fields become `Option`? the Spotify doc mentions nothing ?

Rigellute:
I wasn't able to pindown exactly which properties are now optional - annoyingly this differs on a per track basis.

I counted 19 properties that now need to be optional but then gave up and just made everything optional to be sure to handle other tracks safely.

Not sure why their API has started ommitting properties altogether, seems to be a regression on their end.

ramsayleung:
> I wasn't able to pindown exactly which properties are now optional - annoyingly this differs on a per track basis (or so it appears).

In my honest option, I think we should ask the [upstream](https://github.com/spotify/web-api) to annotate which properties will be optional now instead of just simply making all field optional. It's the duty of upstream to clarify the different cases.

Rigellute:
Looks like this exact issue has already been reported on that repo https://github.com/spotify/web-api/issues/1555

Should we close this now?

ramsayleung:
> Should we close this now?

Do you mean this PR? I think we should let it alone, just waiting for the response from upstream.

ramsayleung:
The upstream has fixed this problem, I will take time to update `rspotify`(probably this weekend) :)

Rigellute:
Great stuff @ramsayleung!

ramsayleung:
The Spotify team has fix the problem
```rust
cargo run --example audio_analysis # parse response successfully.
```
the CI has passed in this [PR](https://github.com/ramsayleung/rspotify/actions/runs/113690526)(but the tests failed sometimes) and I am gonna to close this PR.

By the way, I just added a PR to fix lint errors and make tests more stable, but I occurred some problems, if you are interested in it, you could check this PR for more detail: https://github.com/ramsayleung/rspotify/pull/95 :)

Rigellute:
Thanks @ramsayleung, will close this now.

Will also have a look at your #95

PR #102, opened 2020-07-13
Slightly better documentation, comment formatting, proposal for more changes

marioortizmanero:
Hi @ramsayleung!

I was going to use this library for one of my projects and I've decided I'm going to help a bit. I've formatted consistently and correctly the comments for each endpoint with a vim macro, and have made very small changes to the docs, hoping they're more clear now.

I would love to run a `cargo fmt` to clean everything up automatically, but I understand it'd be better to first merge the two other PRs pending to avoid messing up the diffs.

Another thing I really want to change is the repetition between `blocking/cilent.rs` and `client.rs`. A macro would come in perfectly here to unify blocking and async and I think would make it super easy to add/update new endpoints. I'll make a new PR if you're okay with that (if I manage to get it working). There's got to be a solution for that already anyway, I'll have to investigate. I should open a new issue for that, right?

Edit: I've seen that there's a CI run for cargo fmt, which is failing. Should I just run it in a new commit, then?

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1328,7 +1328,7 @@ impl Spotify {
         &self,
         locale: Option<String>,
         country: Option<Country>,
-        limit: L,
+            limit: L,
         offset: O,
     ) -> Result<PageCategory, failure::Error> {
         let mut params = HashMap::new();
--- a/src/client.rs
+++ b/src/client.rs
@@ -1328,7 +1328,7 @@ impl Spotify {
         &self,
         locale: Option<String>,
         country: Option<Country>,
-            limit: L,
+        limit: L,
         offset: O,
     ) -> Result<PageCategory, failure::Error> {
         let mut params = HashMap::new();
@@ -1588,13 +1588,13 @@ impl Spotify {
     /// Start/Resume a User’s Playback
     /// Provide a `context_uri` to start playback or a album,
     /// artist, or playlist.
-    /// 
+    ///
     /// Provide a `uris` list to start playback of one or more
     /// tracks.
-    /// 
+    ///
     /// Provide `offset` as {"position": <int>} or {"uri": "<track uri>"}
     /// to start playback at a particular offset.
-    /// 
+    ///
     /// Parameters:
     /// - device_id - device target for playback
     /// - context_uri - spotify context uri to play
--- a/src/client.rs
+++ b/src/client.rs
@@ -9,7 +9,7 @@ use serde::de::Deserialize;
 use serde_json::map::Map;
 use serde_json::Value;
 
-//  built-in battery
+// Built-in battery
 use std::borrow::Cow;
 use std::collections::HashMap;
 use std::fmt;
@@ -209,7 +209,7 @@ impl Spotify {
             ))
         }
     }
-    ///send get request
+    /// Send get request
     async fn get(
         &self,
         url: &str,
@@ -227,11 +227,11 @@ impl Spotify {
         }
     }
 
-    ///send post request
+    /// Send post request
     async fn post(&self, url: &str, payload: &Value) -> Result<String, failure::Error> {
         self.internal_call(Method::POST, url, Some(payload)).await
     }
-    ///send put request
+    /// Send put request
     async fn put(&self, url: &str, payload: &Value) -> Result<String, failure::Error> {
         self.internal_call(Method::PUT, url, Some(payload)).await
     }
@@ -240,10 +240,10 @@ impl Spotify {
         self.internal_call(Method::DELETE, url, Some(payload)).await
     }
 
-    ///[get-track](https://developer.spotify.com/web-api/get-track/)
-    ///returns a single track given the track's ID, URI or URL
-    ///Parameters:
-    ///- track_id - a spotify URI, URL or ID
+    /// [get-track](https://developer.spotify.com/web-api/get-track/)
+    /// returns a single track given the track's ID, URI or URL
+    /// Parameters:
+    /// - track_id - a spotify URI, URL or ID
     pub async fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
         let trid = self.get_id(Type::Track, track_id);
         let url = format!("tracks/{}", trid);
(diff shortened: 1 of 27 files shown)


ramsayleung:
>  I've formatted consistently and correctly the comments for each endpoint with a vim macro, and have made very small changes to the docs, hoping they're more clear now.

It's great that you are helping to polish this crate.

> I would love to run a `cargo fmt` to clean everything up automatically.

So do I, and I have set up the Github Action to run `cargo fmt` on a CI, perhaps you could this before you make some PRs.

> Another thing I really want to change is the repetition between `blocking/cilent.rs` and `client.rs`

Sorry for the code smell, it will be great to see only the sole endpoint instead of two almost same repetitions. When I started to refactor this crate to add `async/await` support, I could not find a proper way to unify blocking and async endpoint currently, finally I chose the clumsy but useful solution, just maintaining two versions of endpoint. It will be awesome if you could find a way to clean it up.

> Edit: I've seen that there's a CI run for cargo fmt, which is failing. Should I just run it in a new commit, then?

I think you should fix the errors analyzed by `cargo fmt`, and create a new commit to trigger the CI.

src/client.rs
```diff
@@ -1313,23 +1312,23 @@ impl Spotify {
...
+    /// Minimum: 1. Maximum: 50
+    /// - offset - The index of the first item to return. Default: 0
+    /// (the first object). Use with limit to get the next set of
+    /// items.
     pub async fn categories<L: Into<Option<u32>>, O: Into<Option<u32>>>(
         &self,
         locale: Option<String>,
         country: Option<Country>,
-        limit: L,
+            limit: L,
```
ramsayleung, line 1331: `cargo fmt` complains about the modification :(

marioortizmanero:
> Sorry for the code smell, it will be great to see only the sole endpoint instead of two almost same repetitions. When I started to refactor this crate to add async/await support, I could not find a proper way to unify blocking and async endpoint currently, finally I chose the clumsy but useful solution, just maintaining two versions of endpoint. It will be awesome if you could find a way to clean it up.

I will investigate a bit about how this can be done. Another library called [tekore](https://github.com/felix-hilden/tekore) does this, so I might take inspiration. Ideally, the library used for the requests (`requests` in Python being the most commonly used one) should alternate between blocking and async with a `feature` that can be toggled for `rspotify` as well (`httpx` in Python does this). But on Rust it might be more complicated.

I fixed the formatting issues, you can merge now if you want.

marioortizmanero:
I've tried to remove the repetition in `blocking` at https://github.com/marioortizmanero/rspotify/tree/blocking-rewrite but I'm honestly not experienced enough with async/tokio yet. I followed what `reqwest` does for their `blocking` module, and I wrapped the async implementation in the `blocking::Spotify` struct, executing its methods synchronously by using [a single-threaded runtime](https://tokio-rs.github.io/tokio/doc/tokio/runtime/struct.Builder.html#method.basic_scheduler).

I planned on storing said runtime inside the struct itself, but it's quite messy because the runtime is needed outside of this struct, too, for other implementations like `SpotifyOAuth`. So a global static one would probably be a better idea, although it also requires to be mutable so it requires unsafe code.

Not sure what to do... Any thoughts?

ramsayleung:
>  I fixed the formatting issues, you can merge now if you want.

Pleasure to do that :)

ramsayleung:
> I initially followed what `reqwest` does for their blocking module, and I wrapped the async implementation in the blocking::Spotify struct, executing its methods synchronously by using a single-threaded runtime.

To be honest, I don't think it's a great solution for this case, it's doesn't make a big difference between this with the old version, and we have to handle the `async/tokio` runtime in the blocking `client`, which actually has nothing to do with `tokio` runtime. My intuition is using macro to wrapping them up, but the complex part of this implementation may be the function signature, we need to deal with `async fn` and `fn`.

marioortizmanero:
> To be honest, I don't think it's a great solution for this case, it's doesn't make a big difference between this with the old version, and we have to handle the async/tokio runtime in the blocking client, which actually has nothing to do with tokio runtime.

How does it not make a big difference? The implementation would only be at the async module, and the rest of the structs and functions could be taken from there as well. The thing is that with a procedural macro it seems quite complicated to cover all edge cases, I'm not sure if it's as easy as removing all the `await` found and changing `async fn` into `fn`.

The macro would have to turn this:

```rust
impl Spotify {
    /// Documentation
    #[blocking]
    pub async fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
        // ...
    }
}
```

Into this:

```rust
impl Spotify {
    /// Documentation
    pub async fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
        // ...
    }
}

#[cfg(feature = "blocking")]
mod blocking {
    impl Spotify {
        /// Documentation
        pub fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
            // (awaits removed)
        }
    }
}
```

The runtime could instead turn the original snippet into this:

```rust
impl Spotify {
    /// Documentation
    pub async fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
        // ...
    }
}

#[cfg(feature = "blocking")]
mod blocking {
    impl Spotify {
        /// Documentation
        pub fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
            runtime.block_on(self.spotify.track(track_id));
        }
    }
}
```

Which at first looks more solid because it fully converts async into sync, including any edge cases. And the macro is optional in this case, just to avoid repetition. It could also reduce the binary size because there'd be less code, but I'm not fully sure about that.

The third option I've thought of is that the macro acts as a toggle, rather than having two separate `Spotify`. If `blocking` is enabled, it will be compiled without the `.await`. Otherwise, the original code is maintained, and no `blocking` module is included at all. I can't see why someone would need both async and sync clients, and this would cover the binary size problem I mentioned:

```rust
impl Spotify {
    #[cfg(not(feature = "blocking"))]
    /// Documentation
    pub async fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
        // ...
    }

    #[cfg(feature = "blocking")]
    /// Documentation
    pub fn track(&self, track_id: &str) -> Result<FullTrack, failure::Error> {
        runtime.block_on(self.spotify.track(track_id));
    }
}
```

ramsayleung:
> How does it not make a big difference? The implementation would only be at the async module, and the rest of the structs and functions could be taken from there as well.

Yes, I can get your point now, we can maintain only core codebase of `async`version, and then implement `blocking` version based on `async` version module, instead of just maintaining two versions of copy-and-paste code, what is pretty error-prone.

> The third option I've thought of is that the macro acts as a toggle, rather than having two separate Spotify. If blocking is enabled, it will be compiled without the .await. Otherwise, the original code is maintained, and no blocking module is included at all.

This is also my initial thought, but I think we have to deal with the black magic part of macro, which could be messy. 

> And the final idea I'm not even sure would work is providing the async runtime inside this library itself. That way, Spotify could choose between async and blocking when it's initialized.

Any prototype about this though?  I have no idea how to switch between async and blocking when it's initialized, is there a way to switch between `async fn` and `fn`?

PR #105, opened 2020-08-12
Upgraded dependencies

marioortizmanero:
I ran a `cargo upgrade` from `cargo-edit` to update the dependencies used in this module. I strongly suggest integrating [Dependabot](https://dependabot.com/) here so that this is taken care of automatically. It's really easy to use and avoids having to update all dependencies at once.

The only change I had to make so that the examples compiled was upgrading `percent_encoding` to v2.X. I simply followed the instructions indicated [on their repo](https://github.com/servo/rust-url/blob/master/UPGRADING.md#upgrading-from-percent-encoding-1x-to-2x) for the [encoding set this module used](https://docs.diesel.rs/1.4.x/src/percent_encoding/lib.rs.html#143).

On a side note: I think the `base64` dependency may be unnecessary. I don't see where it's exactly being used, and the examples still compile after removing it from `Cargo.toml` and the `extern crate` from `lib.rs`.

Edit: please merge #106 first so that I can update this afterwards.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,21 +11,21 @@ keywords=["spotify","api"]
 edition = "2018"
 
 [dependencies]
-chrono = { version = "0.4", features = ["serde", "rustc-serialize"] }
-dotenv = "0.13.0"
-failure = "0.1"
-lazy_static = "1.0"
-log = "0.4"
-percent-encoding = "1.0.1"
-rand = "0.6.5"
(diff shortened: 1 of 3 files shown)


marioortizmanero:
I've updated this PR after #106 has been merged, @ramsayleung. You can now review this when you have time.

ramsayleung:
> I strongly suggest integrating Dependabot here so that this is taken care of automatically.

Yeah, I would like to take time to figure out how it works, and integrate it later.

PR #76, opened 2020-02-05
Support other tls implementations besides openssl

marcelbuesing:
Goal is to make it easier to cross compile when using this crate.
Basically I feel a big pain point in cross compilation are always libraries such as libopenssl which have to be available for your target architecture. Rust native implementations make cross compilation a lot easier because you can pretty much just rely on cargo without having to worry about missing some particular library.

Before: 
```
cargo build --release --target arm-unknown-linux-gnueabihf                                            
   Compiling openssl-sys v0.9.54
   Compiling futures-util v0.3.3
error: failed to run custom build command for `openssl-sys v0.9.54`

Caused by:
  process didn't exit successfully: `/home/abc/github-thirdparty/rspotify/target/release/build/openssl-sys-47ed6ba4b4c9b64b/build-script-main` (exit code: 101)
--- stdout
cargo:rustc-cfg=const_fn
cargo:rerun-if-env-changed=ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_LIB_DIR
ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_LIB_DIR unset
cargo:rerun-if-env-changed=OPENSSL_LIB_DIR
OPENSSL_LIB_DIR unset
cargo:rerun-if-env-changed=ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_INCLUDE_DIR
ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_INCLUDE_DIR unset
cargo:rerun-if-env-changed=OPENSSL_INCLUDE_DIR
OPENSSL_INCLUDE_DIR unset
cargo:rerun-if-env-changed=ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_DIR
ARM_UNKNOWN_LINUX_GNUEABIHF_OPENSSL_DIR unset
cargo:rerun-if-env-changed=OPENSSL_DIR
OPENSSL_DIR unset
run pkg_config fail: "Cross compilation detected. Use PKG_CONFIG_ALLOW_CROSS=1 to override"

--- stderr
thread 'main' panicked at '

Could not find directory of OpenSSL installation, and this `-sys` crate cannot
proceed without this knowledge. If OpenSSL is installed and this crate had
trouble finding it,  you can set the `OPENSSL_DIR` environment variable for the
compilation process.

Make sure you also have the development packages of openssl installed.
For example, `libssl-dev` on Ubuntu or `openssl-devel` on Fedora.

[...]

PR #106, opened 2020-08-12
Extern crate, Cargo.toml cleanup

marioortizmanero:
`extern crate` is no longer needed after Rust 2018, so I've removed all of them, using `use` in its place.

I also removed some unnecessary crates:
* `base64`: not used at all
* `env_logger`: not used at all
* `url`: not used at all
* `itertools`: only used once in `client.rs`, found a cleaner solution to avoid using `iproduct`

And other small improvements in `Cargo.toml`.

The one thing I don't know how to solve is the `reqwest`-related features. Wouldn't it make sense to just remove these `extern crate`? The `Cargo.toml` will already make sure the correct features for `reqwest` are being used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,7 @@ reqwest-default-tls ={ version = "0.10", features=["json","socks"], optional = t
 reqwest-native-tls ={ version = "0.10", features=["json","socks", "native-tls"], default-features = false, optional = true,  package = "reqwest"}
 reqwest-native-tls-vendored ={ version = "0.10", features=["json","socks", "native-tls-vendored"], default-features = false, optional = true, package = "reqwest" }
 reqwest-rustls-tls ={ version = "0.10", features=["json","socks", "rustls-tls"], default-features = false, optional = true, package = "reqwest" }
-serde = "1.0"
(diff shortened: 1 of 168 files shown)


ramsayleung:
Thanks for your cleanup, it does remove some unnessary dependencies, which will leads to a smaller binary/crate.

> Wouldn't it make sense to just remove these extern crate?

Yes, it makes sense to let `Cargo.toml` to manage the features and crates. If you want to know more about `reqwest`-related features, you could this PR https://github.com/ramsayleung/rspotify/pull/76, which tried to cross compile when using this crate. Perhaps you could try to remove these `extern crate reqwest` like other crates, and check whether it will fail in the CI of cross-compilation or not. I could merge this PR after you try.

marioortizmanero:
Managed to get it working! Check out the changes in `Cargo.toml`. Only the occasional tests fail now.

ramsayleung:
You have modified tons of files, making it painful for me to review all modification, merged finally :)

marioortizmanero:
Thanks a lot! Most of the files modified were done automatically to remove the `extern crate`.

Issue #108, opened 2020-08-12
Reducing rspotify's core dependencies


marioortizmanero:
I've been creating various PRs to clean up some unnecessary dependencies I found and such. Not only does this decrease compilation times, but this may also reduce binary size in case the optimizer is unable to know a module isn't used at all.

Here are some of my proposals, don't hesitate on criticizing them; I want to discuss these changes with @ramsayleung and if possible with more users of rspotify. It might also be too much to remove some of them:

## Optional features
* Make `webbrowser` optional, under some `cli` feature. This dependency isn't used by all the users, since some may not need to open a browser for the authentication process. It's only used for the `util::request_token` function currently.
* Make `dotenv` optional, under some `env` feature. This is also only used by some users, mostly web servers.

These changes will break backward compatibility, so a new version would have to be released. I could create an `UPGRADING.md` file that explains what changes were made in order to make this process easier.

## Removing dependencies
+ **[accepted PR]** 2 less dependencies after #100 
+ **[pending PR accept]** 4 less dependencies after #106
+ **[pending PR accept]** upgrading the dependencies curiously goes from 191 dependencies with `cargo run --example current_playing` to 177, so ~14 less dependencies after #105
+ **[discussion]** `lazy_static` is only used in `client.rs` for `CLIENT` [here](https://github.com/ramsayleung/rspotify/blob/master/src/client.rs#L40). This `CLIENT` variable is then only used *once* [here](https://github.com/ramsayleung/rspotify/blob/master/src/client.rs#L185). I haven't used `lazy_static` myself, but I wanted to re-evaluate the following questions:
        - Is it necessary that `CLIENT` is a *global, static, public* variable?
        - Is `lazy_static` is necessary here?
+ **[discussion]** Is `rand` necessary only for `util::generate_random_string` [here](https://github.com/ramsayleung/rspotify/blob/master/src/util.rs#L20)? `rand` is a considerably big dependency because it includes lots of things that are often unnecessary. I recently saw this PR https://github.com/rust-analyzer/rust-analyzer/pull/5574 that replaced it with `oorandom`, a much simpler and straightforward crate. Could this also be applied to this use case?

ramsayleung:
> Make `webbrowser` optional, under some cli feature. 

It's a great suggestion, which makes it more easier to build automatic tests without `webbrowser`. But  when we need credential to call Spotify's endpoints, how to get credential without pulling up a browser? It's the original reason of why I used `webbrowser`.

[...]

PR #95, opened 2020-05-24
stablize occasional failed tests

ramsayleung:
Trying to fix failed tests.

PR #110, opened 2020-08-14
Reduce core dependencies

ramsayleung:
Reducing core dependencies, check this issue https://github.com/ramsayleung/rspotify/issues/108 for more details:

- [x] replacing `rand` with a lightweight replacement `getrandom`
- [x] optional `dotenv` dependency
- [x] optional `browser` dependency
- [x] documentation


src/blocking/util.rs
```diff
@@ -12,17 +10,24 @@ use std::string::ToString;
 
 use super::oauth2::{SpotifyOAuth, TokenInfo};
 
+use getrandom::getrandom;
+
+static ALPHANUM: &[u8] =
```
marioortizmanero, line 15: Shouldn't this be inside the `generate_random_string` function? That way we avoid unnecessary global variables.

src/util.rs
```diff
@@ -10,6 +8,10 @@ use std::io;
 use std::string::ToString;
 
 use super::oauth2::{SpotifyOAuth, TokenInfo};
+use getrandom::getrandom;
+
+static ALPHANUM: &[u8] =
```
marioortizmanero, line 13: same for this one

marioortizmanero:
By the way, I don't seem to have permissions to edit this PR. That, or I don't know how to do it:

```
/tmp/rspotify reduce-core-dependencies ⇡ 7s
❯ git push
Username for 'https://github.com': marioortizmanero
Password for 'https://marioortizmanero@github.com': 
remote: Permission to ramsayleung/rspotify.git denied to marioortizmanero.
fatal: unable to access 'https://github.com/ramsayleung/rspotify.git/': The requested URL returned error: 403
```

I think I either have to be a collaborator, or I would have to create the PR, and then you edit it as the original creator.

ramsayleung:
> By the way, I don't seem to have permissions to edit this PR. That, or I don't know how to do it:
> 
> ```
> /tmp/rspotify reduce-core-dependencies ⇡ 7s
> ❯ git push
> Username for 'https://github.com': marioortizmanero
> Password for 'https://marioortizmanero@github.com': 
> remote: Permission to ramsayleung/rspotify.git denied to marioortizmanero.
> fatal: unable to access 'https://github.com/ramsayleung/rspotify.git/': The requested URL returned error: 403
> ```
> 
> I think I either have to be a collaborator, or I would have to create the PR, and then you edit it as the original creator.

My bad, I have invited you as a collaborator.

src/blocking/util.rs
ramsayleung: Yes, it makes sense.

marioortizmanero:
Ok so I've made both of these optional features. The thing is, should any of them be included by default? I don't see `env-file` being default, but `browser` could. Thoughts?

ramsayleung:
> The thing is, should any of them be included by default? I don't see env-file being default, but browser could. Thoughts? 

Agree, I think the `browser` feature should be included by default for which `rspotify` has no other way to help users get their credentials.
```rust
    use std::io;
    match spotify_oauth.get_cached_token() {
        Some(token_info) => Some(token_info),
        None => {
            request_token(spotify_oauth);
            println!("Enter the URL you were redirected to: ");
            let mut input = String::new();
            match io::stdin().read_line(&mut input) {
                Ok(_) => process_token(spotify_oauth, &mut input),
                Err(_) => None,
            }
        }
    }

```

When there is no cache token exists, e.g `spotify_oauth.get_cached_token() ` return false, the only way to authenticate is pulling up a browser. If we have another way to authenticate without pulling up a browser, we could make `browser` feature optional, even remove this feature.

ramsayleung:
Besides, of reducing core dependencies, I also plan to reduce some [`examples`](https://github.com/ramsayleung/rspotify/tree/master/examples) file. All of these files come from a same template, the only difference between them is that they calls different function with different day, everything else is same. I just use example file to test endpoint, but now I think maintain tons of example files is painful, so I prefer to remove some of them to keep this repository clear and maintainable.

marioortizmanero:
Agreed! The examples are quite repetitive and don't offer much help for real projects. I would add more advanced examples instead of lots of very simple ones.

ramsayleung:
I just open a new PR to reduce examples https://github.com/ramsayleung/rspotify/pull/113, feel free to contribute to it.
PS: I think we should merged the PR of reducing examples before this PR, so we don't have to fix those check errors detected by CI.

marioortizmanero:
> PS: I think we should merged the PR of reducing examples before this PR, so we don't have to fix those check errors detected by CI.

Shouldn't we work on #95 first then? I'm not really sure why the tests fail, did you figure something out?

ramsayleung:
> Shouldn't we work on #95 first then? Did you figure something out? The error thread 'test_artist_top_tracks' panicked at 'dispatch dropped without returning error' is quite cryptic, maybe it could also be improved for future users that see it. This issue seems to point out that it's happening because our client is global: hyperium/hyper#2112

Yes, we should. The problem of tests is that the current test cases mostly depend on `Spotify`'s API, the test case would actually send a request to `Spotify`'s server. So the failed tests are pretty mysterious, it's not only the `test_artist_top_tracks` case, but also other cases:
https://github.com/ramsayleung/rspotify/runs/985549966

My intuition about fixing failed tests is just running them more times, for example, put a test case into a loop, and run it repeatedly(e.g 10 times), if any of them passes, the test passes. But it perhaps triggers the [rate-limit strategy](https://developer.spotify.com/documentation/web-api/) of `Spotify` API

marioortizmanero:
Are you sure it's because of the Spotify rate limits anyway? Other Spotify API libraries like [tekore have a 100% code coverage](https://github.com/felix-hilden/tekore/tree/master/tests), meaning that they have even more tests than us, yet I don't recall them dealing with these issues. According to the issue I linked in the previous comment, it might have to do with `hyper`'s client.

ramsayleung:
> Are you sure it's because of the Spotify rate limits anyway? 

To be honest, not sure, just my worry.

> According to the issue I linked in the previous comment, it might have to do with hyper's client.

No clue about this issue.

marioortizmanero:
I think this PR is ready to be merged now as well. Check out the changelog and documentation I wrote and let me know if you'd change anything.

Should `env-file` be enabled for examples like `me` or `album`?

I also fixed a typo in `user_playlist_remove_specific_occurrenes_of_tracks` that I saw while reading code.

Can you get rid of the [GitGuardian security check](https://github.com/ramsayleung/rspotify/pull/110/checks?check_run_id=1037613262)? You have to log into GitGuardian as the owner of the repo and disable it.

ramsayleung:
Merged :)

PR #113, opened 2020-08-16
reduce repeated examples.

ramsayleung:
All of these example files come from a same template, which are are quite repetitive and don't offer much help for real projects, the only difference between them is that they calls different function with different day, everything else is same.

- [x] remove repetitive examples.
- [ ] add some real examples.
- [x] update document.


marioortizmanero:
I would reuse the examples and turn them into tests, rather than completely remove them. That way we have full coverage for every endpoint.

ramsayleung:
> I wanted to suggest to use a refresh token with all scopes already given for the test_with_oauth. We could use it to avoid user interaction, and that way they can be ran with Travis as well, since they aren't supposed to expire. 

Yep, it's a great idea to run tests, and the key point how to get a non-expire refresh-token. Hopefully you could get the premium account.

> By the way, sorry about the force pushes, the 4 following commits don't change anything. The fifth adds a new example called with_refresh_token. Let me know what you think about it.

The example `called with_refresh_token` is helpful to demonstrate how to refresh token, but I think force push it's not a good idea, perhaps it will overwrite something important.

marioortizmanero:
No, the force push was because I created a commit by mistake and then dropped it later, you can do a rebase and merge and it'll be ok, without overwriting anything.

marioortizmanero:
Great! Neat example with the Rocket web app, I like it. Can this be merged for now, then? We can add more examples in the future.

ramsayleung:
Yep, merged :)

PR #114, opened 2020-08-16
Fix tests

marioortizmanero:
I think the issue with the tests is that Rspotify had a bug for multithreaded requests, which is needed for Rust tests. The `reqwest` client (`CLIENT`) was a global with no thread safety at all (for some reason Rust didn't detect that). Thus, there are two possible solutions:

1. Keep the global client, but wrap it with an `Arc<Mutex<T>>` so that it's thread-safe. But global state is usually a bad idea IMO.
2. Move the client inside the `Spotify` struct itself. This means that it can no longer derive `Serialize` and `Deserialize`, but I don't see why these were needed anyway, and it makes sense to me to have a client for each `Spotify` object.

~~By the way, I would like to know why `Serialize` and `Deserialize` are needed for structs such as `Spotify` and `ApiError`. Shouldn't these be cleaned up?~~ Okay I've just seen this commit https://github.com/ramsayleung/rspotify/commit/b7b7c6f8b959724699eaf656a2b22de881e2e818

This should close #95.

### DIFF
--- a/src/blocking/client.rs
+++ b/src/blocking/client.rs
@@ -2,13 +2,12 @@
 // 3rd-part library
 use chrono::prelude::*;
 use failure::format_err;
-use lazy_static::lazy_static;
 use log::{error, trace};
 use reqwest::blocking::Client;
 use reqwest::header::{HeaderMap, AUTHORIZATION, CONTENT_TYPE};
 use reqwest::Method;
 use reqwest::StatusCode;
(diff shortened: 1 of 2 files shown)


ramsayleung:
Awesome! merged :)

ramsayleung:
From this case, I think we could just avoid Singleton, and make tokio::Runtime as a field of `Struct Spotify` ?

PR #115, opened 2020-08-16
Fix pull requests running CI twice

marioortizmanero:
I noticed that the Continuous Integration was running twice in each PR: one because of a `push`, and another because of a `pull_request`. This should fix that by only running for `push` when it's on `master`.


ramsayleung:
Merged :)

PR #117, opened 2020-08-16
Reducing boilerplate for tests

marioortizmanero:
**test_device:** Not sure what that's for. Can it be removed?
**test_with_credential:** Just a function to reduce the boilerplate a bit.
**test_oauth:** This creates a single spotify client so that it can be used from each test. This way, you don't have to open a browser tab for each test, and there's much less copy-pasted code. This is specially useful in case we figure out how to run these tests on Travis, because it'll be very easy to change the authentication process.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ serde_json = "1.0.57"
 webbrowser = "0.5.5"
 
 [dev-dependencies]
+async_once = { version = "0.1.0", features = ["tokio"] }
 tokio = { version = "0.2.22", features = ["full"] }
 futures = "0.3.5"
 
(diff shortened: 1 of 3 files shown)


ramsayleung:
> test_device: Not sure what that's for. Can it be removed?

Yes, of course, it's useless now.

> This creates a single spotify client so that it can be used from each test. This way, you don't have to open a browser tab for each test, and there's much less copy-pasted code. This is specially useful in case we figure out how to run these tests on Travis, because it'll be very easy to change the authentication process.

Yes, we don't need so much copy-pasted code, it will be more clear to remove them, good to see that.

I prefer to keep this PR on until the Spotify Offical responses to our request about applying for a premium account for the library

marioortizmanero:
> Yes, of course, it's useless now.

Okay, I've removed it.

> I prefer to keep this PR on until the Spotify Offical responses to our request about applying for a premium account for the library

They've flagged my post as spam so I don't think we can get that premium account. If I manage to do so, I'll let you know. But you can merge for now.

Issue #112, opened 2020-08-15
Cleaning up the blocking module


marioortizmanero:
This is a continuation of #102, which discussed how to properly export the `blocking` module without copy-pasting the async functions and removing `async` and `await`. The current solution makes it painful to maintain rspotify, and compilation times when using the blocking client are huge because you're basically compiling rspotify twice, which kinda defeats the point of #108.

I first thought that using an attribute macro could be a good idea because all that needs to be modified from the async implementations is removing the `async` and `await` keywords. See [this comment](https://github.com/ramsayleung/rspotify/pull/102#issuecomment-660632560) for a demonstration. But this still doesn't fix the "compiling the codebase twice" issue, because the code would still be repeated, now after the macro is ran instead of manually. The macro in question would make compilation times even longer, and could be messy to implement. The only possible way to only "compile the codebase once" would be by having the `blocking` feature *only* export the blocking interface when used, but I'm not sure we can assume that the users won't need both the async and blocking interfaces for a program.

The second possible solution is calling the async implementations in the blocking functions using `block_on`. This might sound less efficient at first, but here's a comparison I made to prove my point:

```rust
use std::time::Instant;

async fn original() -> Result<String, reqwest::Error> {
    reqwest::get("https://www.rust-lang.org")
        .await?
        .text()
        .await
}

fn with_copypaste() -> Result<String, reqwest::Error> {
    reqwest::blocking::get("https://www.rust-lang.org")?
        .text()
}

fn with_block_on() -> Result<String, reqwest::Error> {
    let mut rt = tokio::runtime::Builder::new()
        .basic_scheduler()
        .enable_all()
        .build()
        .unwrap();

    rt.block_on(async move {
        original().await
    })
}

fn main() {
    let now = Instant::now();
    print!("benchmarking with copypaste ...");
    for _ in 1..100 {
        with_copypaste().unwrap();
    }
    println!("done in {}ms", now.elapsed().as_millis());

    let now = Instant::now();
    print!("benchmarking with block_on ...");
    for _ in 1..100 {
        with_block_on().unwrap();
    }
    println!("done in {}ms", now.elapsed().as_millis());
}
```

```
❯ cargo run --release
    Finished release [optimized] target(s) in 0.04s
     Running `/home/mario/.cache/cargo/release/macros`
benchmarking with copypaste ...done in 31988ms
benchmarking with block_on ...done in 30250ms
```

[...]

PR #120, opened 2020-08-29
Blocking module cleanup

marioortizmanero:
This is a WIP implementation of #112.

It removes duplicated implementations in the blocking module. I've found three possible cases of how this implementation can be hidden, after creating a wrapper in the blocking module like so:

```rust
#[derive(Debug, Clone, Deref)]
pub struct Spotify(crate::client::Spotify);
```

* **Simple methods that can be dereferenced from the original type**: some can be automatically taken care of by `Deref` and `DerefMut`. For example, `rspotify::oauth2::SpotifyClientCredentials::is_token_expired`. These will dereference the blocking wrapper into the original async struct, and call the async struct's method. These traits are automatically implemented with the `derive_deref` crate, which is quite handy for our use case. This part avoids lots repetitive code but might cause issues with the typing system, since this dereference is only wanted to call the underlying type's methods. Thus, this case could be removed, writing the rest of the underlying methods like so:

```rust
fn is_token_expired(&self, token_info: &TokenInfo) -> bool {
    self.0.is_token_expired(token_info)
}
```

* **Methods that return `Self`**: if this was left to `Deref`, the method would return the *underlying type*, not the blocking wrapper type. For example, `rspotify::blocking::Spotify::default`. Thus, this requires to write the *function signature* again in the blocking module, along with some boilerplate code. This boilerplate code may vary, but it's usually got to do with the builder pattern, and looks something like this:

```rust
pub fn access_token(mut self, access_token: &str) -> Self {
    self.0 = self.0.access_token(access_token);
    self
}
```

* **Async methods**: these methods require applying some specific code in order to run them with a blocking runtime. For example, `rspotify::blocking::Spotify::search`. This can be done in different ways, which is discussed in #112, but for this PR I used the simplest one for now:

```rust
fn with_block_on_handle() -> Result<String, reqwest::Error> {
    RT.handle().block_on(async move {
        original().await
    })
}
```

I'm not sure if the async -> sync conversion with the global runtime can be done automatically with a macro, because some of these methods have parameters, and I'm not sure how that would work in a macro. At most we could have one that does this:

```rust
fn with_block_on_handle() -> Result<String, reqwest::Error> {
    blocking! {
        original()
    }
}
```

Which is still an improvement. NOTE: this is now implemented in the `src/macros.rs` file.

I still have to figure out if the documentation can be copied from the original method.

I only implemented the conversion for the `search` and `me` methods in `Spotify` because I only wanted to show how I think it could be done. These can be tested with the available examples (as a note, the `me` example doesn't work at all, but this might be because either the example or the original implementation are bugged, I'll have to look into it):

```bash
$ cargo run --example search --features="env-file blocking"
```

What do you think?

**TODO**:
- [ ] Choose the best async to sync method
- [ ] Clean up macro, make it as flexlible as possible and document it
- [ ] Hide the macros outside the crate (maybe `[#[doc(hidden)]`?)
- [ ] Does the macro work for generic functions? Make it work for these as well if it doesn't.
- [ ] Might be unnecessary, but the macro could also accept other procedural macros applied to the method.
- [ ] Apply macro to all endpoints in `client.rs`
- [ ] Apply macro to structs in `oauth.rs`
- [ ] Remove `pub(in crate)` from `SpotifyClientCredentials` and `SpotifyOAuth`


ramsayleung:
> Simple methods that can be dereferenced from the original type: some can be automatically taken care of by Deref and DerefMut. For example, rspotify::oauth2::SpotifyClientCredentials::is_token_expired. These will dereference the blocking wrapper into the original async struct, and call the async struct's method

I am a little bit confused why we can dereference the blocking wrapper into the async struct, do you mean the blocking wrapper is something like `BlockingWrapper<AsyncVersionStruct>`, and we can get the async struct by dereferencing the blocking wrapper:
```rust
#[derive(Debug, Clone, Serialize, Deserialize, Deref, DerefMut)]
pub struct SpotifyClientCredentials(pub(in crate) AsyncSpotifyClientCredentials);
```

As for the blocking implementation:
```rust
fn with_block_on_handle() -> Result<String, reqwest::Error> {
    RT.handle().block_on(async move {
        original().await
    })
}
```

the implementation is simple and clear, but will it cause race condition when using a global variable without lock?

> I'm not sure if the async -> sync conversion with the global runtime can be done automatically with a macro, because some of these methods have parameters, and I'm not sure how that would work in a macro

If global runtime can not be done automatically with a macro, perhaps we could change the implementation to use a local runtime inside a function if it can make it easy to do with a macro. I'm not sure how that work in a macro either, to be honest, I am not that familiar with macro(the black magic part).

>I only implemented the conversion for the `search` and `me` methods in Spotify because I only wanted to show how I think it could be done.

Well, I could not address the `me` method, do I miss something? On the other side, the `search` method is much more clearer than before, now it's a true "blocking" method, instead of a copy-paste version.

marioortizmanero:
> I am a little bit confused why we can dereference the blocking wrapper into the async struct, do you mean the blocking wrapper is something like BlockingWrapper<AsyncVersionStruct>, and we can get the async struct by dereferencing the blocking wrapper:

Exactly. This way we automatically have access to all the async's methods.

> the implementation is simple and clear, but will it cause race condition when using a global variable without lock?

No, the [`tokio::runtime::Handle`](https://docs.rs/tokio/0.2.22/tokio/runtime/struct.Handle.html) returned by `RT.handle()` is specifically designed to manage tasks. Using `Arc<Mutex<T>>` is unnecessary in this case, [as it may be moved between threads safely](https://docs.rs/tokio/0.2.22/tokio/runtime/struct.Runtime.html#method.handle) (the tokio docs could definitely be improved, though).

> If global runtime can not be done automatically with a macro, perhaps we could change the implementation to use a local runtime inside a function if it can make it easy to do with a macro. I'm not sure how that work in a macro either, to be honest, I am not that familiar with macro(the black magic part).

You can check out the macro [here](https://github.com/ramsayleung/rspotify/blob/blocking-cleanup/src/macros.rs). Macros are fun and super useful, you should check them out. In this case, I think a function could also be used, I'm not sure what's the better choice.

> Well, I could not address the me method, do I miss something? On the other side, the search method is much more clearer than before, now it's a true "blocking" method, instead of a copy-paste version.

The `me` method can be found [here](https://github.com/ramsayleung/rspotify/blob/blocking-cleanup/src/blocking/client.rs#L767). Try out the example with this command and see how there's an error that points out that the scope is illegal:

```bash
$ cargo run --example search --features="env-file blocking"
```

ramsayleung:
> The me method can be found here. Try out the example with this command and see how there's an error that points out that the scope is illegal.

I can reproduce this error, it turns out that Spotify has changed the required scopes of `me` endpoint, changing the scope to this can fix this error:

```rust
    let mut oauth = SpotifyOAuth::default()
        .scope("user-read-currently-playing playlist-modify-private")
        .build();
```

marioortizmanero:
> I can reproduce this error, it turns out that Spotify has changed the required scopes of me endpoint, changing the scope to this can fix this error:

Can you update that in a new PR?

I will be creating a procedural macro to automatically implement the blocking module with `#[wrap_blocking]` or something like that. I'm sure there are more libraries that may want to use it, so I'll make it available for everyone. I'll continue working on this afterwards.

ramsayleung:
> Can you update that in a new PR?

Yep, I have created a new PR, which will be merged later: https://github.com/ramsayleung/rspotify/pull/121

>  I'm sure there are more libraries that may want to use it, so I'll make it available for everyone. 

Great!!

marioortizmanero:
Never mind! I've managed to create a `macro_rules` one that even duplicates the documentation! Check out the macros.rs and `track` implementation at `client.rs`.

marioortizmanero:
OK so turns out that parsing generic parameters with macro_rules is quite complicated, or almost difficult.

May I know why some types use `Into`, like `Into<Option<Country>>` in `artist_top_tracks`, or `Into<Option<u32>>` in `search @ramsayleung? Couldn't they just use `Option<Country>` and `Option<u32>`, respectively?

ramsayleung:
>Couldn't they just use Option<Country> and Option<u32>, respectively? These also don't seem consistent, as this is only used for some of the Option<T>

Back to the time when still was an undergraduate, I started to develop `rspotify`, I am not that familiar with Rust, I struggled with the default values for function. In my opinion, some parameters in query endpoints could have a default value, such as `Country`, `limit`, `offset`. But I found no way to set a default value for function as what Python and C++ do, so I looked for help in the Rust community, they gave me some suggestions:

https://www.reddit.com/r/rust/comments/7u1zjl/question_about_default_values_for_function/

So,

marioortizmanero:
> Back to the time when still was an undergraduate, I started to develop rspotify, I am not that familiar with Rust, I struggled with the default values for function. In my opinion, some parameters in query endpoints could have a default value, such as Country, limit, offset. But I found no way to set a default value for function as what Python and C++ do, so I looked for help in the Rust community, they gave me some suggestions:

Ahh I get it now. It's just to avoid having to use `Some` in function calls. It is a neat solution but I'm not sure if it's the best one, because the generic parameter will generate multiple copies of the same function (2^N, as [this comment points out](https://www.reddit.com/r/rust/comments/7u1zjl/question_about_default_values_for_function/dthjxqh/)). And as I said, it's a bit unconsistent, because not all optional parameters use that. As always, I'd just look at how other libraries do it and try to follow that. I'll investigate about that.

ramsayleung:
It's this draft ready to create a PR? If it's ready to do so, the PR https://github.com/ramsayleung/rspotify/pull/128 should wait before this PR merged, it's a little bit painful to modify two sets of code at the same time.

marioortizmanero:
No, sorry. Go ahead and finish #128, this one is going to take more time. Let me know before you merge it so that I can review it.

marioortizmanero:
Closing this PR in favor of #129.

Issue #116, opened 2020-08-16
Explicitly read from the environment variables with `from_env`


marioortizmanero:
Calling `SpotifyClientCredentials::default()` [automatically tries to read the environment values for `CLIENT_ID` and `CLIENT_SECRET`](https://github.com/ramsayleung/rspotify/blob/master/src/oauth2.rs#L101). I think this shouldn't be the default behavior, meaning that the current `default` method should become `from_env`, and have a `default` method that simply sets the values to `None`.

This is also done by the [tekore](https://tekore.readthedocs.io/en/latest/reference/config.html?highlight=env#tekore.config_from_environment) library, for example.

This should be done after #109 is taken care of. I have to check if any other structs apart from `SpotifyClientCredentials` do anything similar.

Issue #109, opened 2020-08-13
Using `derive_builder` to avoid repetition with the builder pattern


marioortizmanero:
When I was reading the code in `src/oauth.rs` and `src/client.rs` I couldn't help but notice all the boilerplate and repetition needed to implement builder patterns for some of the structs declared.

I think this would be a great opportunity to use the [`derive_builder`](https://docs.rs/derive_builder/0.9.0/derive_builder/) crate in order to have the codebase more maintainable, easier to read, and consistent. Read more about the features it provides [here](https://github.com/colin-kiegel/rust-derive-builder#usage-and-features).

The downside is possibly longer compilation times, but after improving this part in the latest commits I wouldn't think it's much of a problem. And I would like to compare the compilation times before and after the changes before this is merged into master to see how much of a difference there actually is.

Furthermore, using `From<String>` instead of `&str` for the [builder parameters like this one](https://github.com/ramsayleung/rspotify/blob/master/src/oauth2.rs#L109) would avoid an extra allocation for whenever the user passes a `String`, since the `to_owned()` can be omitted, and it will be directly moved into the struct. If the user passes a `&str` it will just be cloned. This is also supported by [`derive_builder`](https://docs.rs/derive_builder/0.7.1/derive_builder/#generic-setters).

PR #123, opened 2020-09-06
Replace failure with anyhow and thiserror

### LINKED ISSUE #122: Failure is now deprecated
This appears as a direct dependency and as a dependency of an old version of dotenv.

Result of `cargo audit`:

```
warning: 1 warning found

Crate:  failure
Title:  failure is officially deprecated/unmaintained
Date:   2020-05-02
URL:    https://rustsec.org/advisories/RUSTSEC-2020-0036
Dependency tree: 
failure 0.1.8
├── rspotify 0.10.0
│   └── holding 0.1.0
└── dotenv 0.13.0
    └── rspotify 0.10.0
```

arlyon:
As described in #122 , `failure` is deprecated. This replaces it with anyhow and thiserror.

Closes #122

### DIFF
--- a/src/blocking/client.rs
+++ b/src/blocking/client.rs
@@ -1,20 +1,18 @@
 //! Client to Spotify API endpoint
 // 3rd-part library
 use chrono::prelude::*;
-use failure::format_err;
 use log::{error, trace};
-use reqwest::blocking::Client;
+use reqwest::blocking::{Client, Response};
 use reqwest::header::{HeaderMap, AUTHORIZATION, CONTENT_TYPE};
-use reqwest::Method;
-use reqwest::StatusCode;
+use reqwest::{Method, StatusCode};
 use serde::Deserialize;
 use serde_json::map::Map;
 use serde_json::{json, Value};
+use thiserror::Error;
 
 //  Built-in battery
 use std::borrow::Cow;
 use std::collections::HashMap;
-use std::fmt;
 use std::io::Read;
 use std::string::String;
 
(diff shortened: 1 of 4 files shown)


marioortizmanero:
Thanks a lot for the PR!

I've been thinking if failure is necessary since I first saw it in the codebase. Do you think anyhow is worth it still? I read that these kind of libraries are often used for binaries, not libraries.

This is a good opportunity to just get rid of failure/anyhow. I'll investigate a bit about that when I have time.

arlyon:
Hey, yeah I seem to have the same thought. Ideally we'd only use `thiserror` so that people using the code can error handle better but I figured this would be a good first step :)

marioortizmanero:
I think we could remove `anyhow` in this same PR, if you're up for it. I like the `Result` type alias because it makes the endpoints more concise, so we could use an internal `Result` type as well (`type Result = std::result::Result<T, ApiError>`)

 I don't personally use `thiserror` myself because I consider it unnecessary, but it is indeed quite nice. I would like an opinion from @ramsayleung, who is the main maintainer.

ramsayleung:
> I think we could remove anyhow in this same PR, if you're up for it.

I kind of agree with @marioortizmanero, `anyhow` is seems not that necessary, we could use an internal `Result` type as what Mario suggests: `type Result = std::result::Result<T, ApiError>`.

Oh the other hand, `thiserror`'s doc also points out:

> Use Anyhow if you don't care what error type your functions return, you just want it to be easy. This is common in application-like code.

As for `thiserror`, it is nice and acceptable, we could use it to reduce some hand-write boilerplate code.

arlyon:
Sounds good to me. If you're both behind it I'll see what I can do!

arlyon:
I've added some changes. Please have a look at the new error type and let me know what you think. Tried to follow `thiserror` convention when possible.

arlyon:
I also have a branch with a few 'misc' changes like some clippy lints and code fixes, should I open another PR for those?

https://github.com/ramsayleung/rspotify/compare/master...arlyon:misc-fixes

ramsayleung:
> I also have a branch with a few 'misc' changes like some clippy lints and code fixes that I came across while working on this, should I open another PR for those?

Yep, it would be better to open a new PR for these 'misc' changes, because there is nothing related to this PR: `anyhow` and `thiserror`. It could be much clear to open a new PR to illustrate what you wan to do :)

src/blocking/client.rs
```diff
@@ -41,74 +39,70 @@ use crate::senum::{
...
+}
+
+impl From<Response> for ClientError {
+    fn from(response: Response) -> Self {
+        match response.status() {
+            StatusCode::UNAUTHORIZED => Self::UnauthorizedError,
+            StatusCode::TOO_MANY_REQUESTS => Self::RateLimitedError(
+                response
+                    .headers()
+                    .get(reqwest::header::RETRY_AFTER)
```
arlyon, line 66: This was rewritten because indexing directly can panic (though it is unlikely). Doing a match on `response.headers().get(...).and_then(|header| header.to_str().ok())` was suddenly a lot going on so I used another `and_then` instead.

src/blocking/client.rs
```diff
@@ -41,74 +39,70 @@ use crate::senum::{
...
+                    .and_then(|header| header.to_str().ok())
+                    .and_then(|duration| duration.parse().ok()),
+            ),
+            status @ StatusCode::FORBIDDEN | status @ StatusCode::NOT_FOUND => response
+                .json::<ApiError>()
+                .map(Into::into)
+                .unwrap_or_else(|_| status.into()),
```
arlyon, line 73: This was rewritten simply to match the same style as above.

src/blocking/client.rs
marioortizmanero: I agree, It's better this way IMO

marioortizmanero:
Looks good to me! The only nitpick: IMO the "Error" suffix in the error types are unnecessary. `ClientError` already points out that its children are different types of errors. `Unauthorized` for instance would be enough on its own, because otherwise you have to use `ClientError::UnauthorizedError`. Thoughts on this?

Can you also add this in the `CHANGELOG.md`?

src/blocking/client.rs
ramsayleung: Yeah, it's more robust than before :0

arlyon:
Its a question of naming style suppose. `thiserror` doesn't use it in their examples (which I agree is nicer) while the [rust api guidelines](https://rust-lang.github.io/api-guidelines/naming.html#names-use-a-consistent-word-order-c-word-order) keeps it.

I'll update the changelog

marioortizmanero:
Thanks for your help! This can probably be merged now (not sure if I should be merging PRs myself).

@ramsayleung, should we think about releasing a new version, or do we wait for #120? That one is going to take longer than I expected and there are quite a few breaking changes already. Or maybe it's better to release all the breaking changes at once? Not sure.

ramsayleung:
> should we think about releasing a new version, or do we wait for #120?

I suggest to wait for #120, so we don't need to release a version of breake changes twice. 

> We could create a new issue tracking the new release like many other libraries do, or at least use GitHub milestones.

Yep, It's a good idea. I personnally suggest to create a milestone of stable version, and label this version as the one last version of breake change before the stable version.

After cleaning up the blocking module, and refacting codebase of error handle, I think we are marching to the stable version, `rspotify` will be stable enough to reach `version 1.0`

ramsayleung:
Merged :)

marioortizmanero:
> Yep, It's a good idea. I personnally suggest to create a milestone of stable version, and label this version as the one last version of breake change before the stable version.
>
> After cleaning up the blocking module, and refacting codebase of error handle, I think we are marching to the stable version, rspotify will be stable enough to reach version 1.0

I personally think we're nowhere near a stable version yet. After #120, there's still #116 and #109 with breaking changes. #4 might also include some breaking changes, and the newly opened #124 could change some things as well.

ramsayleung:
Ooops, it still have some features needing to be implemented :)

Issue #127, opened 2020-09-13
Meta-Issue


Kestrer:
This is a meta-issue of many issues with rspotify. Some of it's opinionated but most of it I think is just an improvement.

### General

- [x] #173 The crate root contains nothing other than modules. This means that paths get very long and verbose (e.g. `rspotify::client::Spotify`). The entire contents of the `client` module should be moved to the crate root to make it easier to access.
- [x] #128 `senum`; this module contains all the enums used (and also `Error` for some reason), but could be split into more fitting modules such as `model`.
- [x] #129 Rspotify is designed for use from a CLI. Lots of the functions read from stdin, but this is out of scope for this library. Either they should be removed or contained in a `cli`/`tui`/`terminal`/`console` module and feature-gated.
- [x] #129 This library extensively uses a fake builder pattern - it's like the real builder pattern but it's all the same type. To add proper type safety the builder patterns should be properly implemented in their own types.
- [x] #128 Many functions like `Spotify::tracks` take in a `Vec<&str>`/`Vec<String>`/`&[String]`/`&[&str]`, when they should take in an `impl IntoIterator<Item = &str>`.
- [x] #128 Some endpoints and functions in Rspotify take `String` as parameters instead of `&str`.
- [x] ~~`limit` and `offset` parameters should universally be `usize`s as they are easier to work with from Rust.~~
- [x] #129 Actual `Default::default impl` (see `TokenInfo::default`)
- [x] #129 `Spotify::get_uri` should use`format!`:

```rust
    fn get_uri(&self, _type: Type, _id: &str) -> String {
        format!("spotify:{}:{}", _type.as_str(), self.get_id(_type, _id))
    }
```
- [x] #129 Fix the documentation links to Spotify endpoints and make the format more consistent
- [x] #161 Make a separate type for IDs parsed by `Spotify::get_id`
- [x] #226 Authenticated tests should actually not modify the user's account and cause minimal changes (as in, don't completely delete the user's library for a test).

### OAuth

- [x] #185  `TokenInfo` stores `expires_in` as a `u32`, and additionally has the `expires_at` field. This is not easy to use, instead it should have a single `expires` field with the type `Instant`.
- [x] #185 `TokenInfo` stores a `scope: String`, when it should store a `scope: Vec<String>` obtained by splitting the scopes by spaces.
- [x] #129 `TokenInfo` stores `token_type`, which is useless since it is always `Bearer`.
[...]

PR #128, opened 2020-09-19
Polish Rspotify

ramsayleung:
Clean up Rspotify, take some points of https://github.com/ramsayleung/rspotify/issues/127 into consideration.
+ [x] ~~remove convert_map_to_string and convert_string_to_map in `util.rs`, replaced them with `Reqwest's` `query` method and `Url` struct.~~
+ [x] Many functions like Spotify::tracks take in a Vec<&str>/Vec<String>/&[String]/&[&str], when they should take in an impl IntoIterator<Item = &str>.
+ [x] Split single `senum.rs` file into a module named `enums` and move it into `model` module.
+ [x] reexport struct in `model` module to allow users to write `rspotify::model::FullAlbum` instead of the overly verbose `rspotify::model::album::FullAlbum`.
+ [x] Rename CurrentlyPlaybackContext to CurrentPlaybackContext.
+ [x] Change release_date_precision to enum.


src/client.rs
```diff
@@ -244,7 +232,7 @@ impl Spotify {
     /// - market - an ISO 3166-1 alpha-2 country code.
     pub async fn tracks(
         &self,
-        track_ids: Vec<&str>,
+        track_ids: impl IntoIterator<Item = &str>,
```
marioortizmanero, line 235: One thing I've realized is the inconsistency with `impl IntoIterator<Item = &str>`. The changelog indicates that this is the new function signature for iterable parameters, but sometimes you use `impl IntoIterator<Item = String>` or `impl IntoIterator<Item = &String>`. I would go with the `&str` solution to avoid unnecessary allocations, and only use that type in this module.

src/oauth2.rs
```diff
@@ -379,13 +378,14 @@ impl SpotifyOAuth {
...
-        let query_str = convert_map_to_string(&payload);
-        let mut authorize_url = String::from("https://accounts.spotify.com/authorize?");
-        authorize_url
-            .push_str(&utf8_percent_encode(&query_str, PATH_SEGMENT_ENCODE_SET).to_string());
-        trace!("{:?}", &authorize_url);
-        authorize_url
+        Url::parse_with_params(AUTHORIZE_URL, payload.clone())
+            .unwrap_or_else(|_| {
+                panic!(
+                    "Fail to generate authorize_url, url: {}, payload: {:?}",
```
marioortizmanero, line 384: This should be "Failed", not "Fail"

CHANGELOG.md
```diff
@@ -11,7 +11,10 @@
...
   + Cleanup, reduced repetitive code and boilerplate internally in several places ([#117](https://github.com/ramsayleung/rspotify/pull/117), [#113](https://github.com/ramsayleung/rspotify/pull/113), [#107](https://github.com/ramsayleung/rspotify/pull/107), [#106](https://github.com/ramsayleung/rspotify/pull/106)).
-- Updated dependencies to the latest versions, integrated Dependabot to keep track of them ([#105](https://github.com/ramsayleung/rspotify/pull/105), [#111](https://github.com/ramsayleung/rspotify/pull/111)).
+  + Updated dependencies to the latest versions, integrated Dependabot to keep track of them ([#105](https://github.com/ramsayleung/rspotify/pull/105), [#111](https://github.com/ramsayleung/rspotify/pull/111)).
+  + Endpoints take `Vec<String>/[String]` as parameter have changed to `impl IntoIterator<Item = &str>`, which is backward compatibility.
```
marioortizmanero, line 15: IMO this should be:

```
Endpoints that take `Vec<String>/[String]` as a parameter have changed to `impl IntoIterator<Item = &str>`.
```

And just be left outside the "Breaking changes" section.

src/client.rs
ramsayleung: > One thing I've realized is the inconsistency with impl IntoIterator<Item = &str>. The changelog indicates that this is the new function signature for iterable parameters

Yes, I have noticed the `inconsistency` with `impl IntoIterator<Item = &str>`, but it's caused by the original design, not the introduction of `impl IntoIterator<Item = &str>`. 

> but sometimes you use impl IntoIterator<Item = String> or impl IntoIterator<Item = &String>

Because some functions use `&[String]` and some functions use `Vec<&str>`, If I change them to `impl IntoIterator<Item = &String>`, it will break up the backward compatibility.

> I would go with the &str solution to avoid unnecessary allocations, and only use that type in this module.

I have thinked about this solution, it's fine to break up backward compatibility at this release.

marioortizmanero: I would say it's OK to break backward compatibility with this. We're going to change lots of other things anyway.

ramsayleung:
@marioortizmanero I think this PR is ready to review, as for the rest of issues mention in #127 which are mostly about Model, I prefer to create a new PR, otherwise this PR will be a huge one which is difficult to review.

marioortizmanero:
The thing is that I'm doing *tons* of changes at #129, specially related to the endpoints, so the merge conflict is going to be huge if we merge this right now. Can you remove the changes you've done to the endpoints and OAuth to avoid this? We can take care of them after it's merged.

Very sorry for the inconvenience. I thought it was going to require less changes but I was forced to even rewrite the OAuth part. The endpoints work differently now as well, so I had to change them as well.

ramsayleung:
>  #129 is almost ready as well, can you start taking a look in there? If you can help me to update the docs and tests it'd be super useful.

Yeah, Sure. I am gonna to review thie PR, and collaborate to update docs and tests :)

src/blocking/client.rs
```diff
@@ -180,15 +179,14 @@ impl Spotify {
...
+                .request(method.clone(), &url.into_owned())
                 .headers(headers);
 
-            // Only add body if necessary
-            // spotify rejects GET requests that have a body with a 400 response
-            let builder = if let Some(json) = payload {
-                builder.json(json)
-            } else {
-                builder
+            let builder = match method {
```
ramsayleung, line 185: @marioortizmanero Check this function for more details.

marioortizmanero:
The diff is all messed up for this PR. Is there a way to remove my commits from the PR we just merged so that I can review this?

src/client.rs
```diff
@@ -11,29 +11,29 @@ use thiserror::Error;
...
-use super::model::context::{CurrentlyPlaybackContext, CurrentlyPlayingContext};
-use super::model::cud_result::CUDResult;
-use super::model::device::DevicePayload;
-use super::model::page::{CursorBasedPage, Page};
-use super::model::playing::{PlayHistory, Playing};
-use super::model::playlist::{FeaturedPlaylists, FullPlaylist, PlaylistTrack, SimplifiedPlaylist};
-use super::model::recommend::Recommendations;
-use super::model::search::SearchResult;
-use super::model::show::{
+use super::model::CUDResult;
```
marioortizmanero, line 19: I think this is one of the few acceptable moments where you can use `use super::model::*` and `use super::enums::*`. The client should be using all of them anyway, and it cleans up the header a lot. What do you think? That, or merging all the `use super::model` into a huge single import. I prefer the first option, though.

CHANGELOG.md
```diff
@@ -12,8 +12,10 @@ If we missed any change or there's something you'd like to discuss about this ve
   + Remove `itertools` dependency by using the standard library.
   + Remove `rand` in place of `getrandom` to [reduce total dependencies and compile times](https://github.com/ramsayleung/rspotify/issues/108#issuecomment-673587185).
   + Cleanup, reduced repetitive code and boilerplate internally in several places ([#117](https://github.com/ramsayleung/rspotify/pull/117), [#113](https://github.com/ramsayleung/rspotify/pull/113), [#107](https://github.com/ramsayleung/rspotify/pull/107), [#106](https://github.com/ramsayleung/rspotify/pull/106)).
-- Updated dependencies to the latest versions, integrated Dependabot to keep track of them ([#105](https://github.com/ramsayleung/rspotify/pull/105), [#111](https://github.com/ramsayleung/rspotify/pull/111)).
-
+  + Updated dependencies to the latest versions, integrated Dependabot to keep track of them ([#105](https://github.com/ramsayleung/rspotify/pull/105), [#111](https://github.com/ramsayleung/rspotify/pull/111)).
++ Remove `convert_map_to_str` and `convert_str_to_map` from both `util.rs` and `blocking/util.rs`, replacing them with `reqwest`'s `query` and `Url::Parse`. Remove crate `percent-encoding`
```
marioortizmanero, line 16: This can be removed now.

CHANGELOG.md
```diff
@@ -12,8 +12,10 @@ If we missed any change or there's something you'd like to discuss about this ve
...
++ Remove `generate_random_string` and `datetime_to_timestamp` from `blocking/util.rs`, using `generate_random_string` and `datetime_to_timestamp` from `util.rs` instead.
```
marioortizmanero, line 17: This can be removed now.

CHANGELOG.md
```diff
@@ -38,6 +40,33 @@ If we missed any change or there's something you'd like to discuss about this ve
...
+	- `current_user_saved_albums_delete`
+	- `current_user_saved_albums_contains`
+	- `user_follow_artists`
+	- `user_unfollow_artists`
+	- `user_artist_check_follow`
+	- `user_follow_users`
+	- `user_unfollow_users`
+	- `audios_features`
+- Reexport `model` module to allow user to write `rspotify::model::FullAlbum` instead of  `rspotify::model::FullAlbum`.
+- Split single `senum.rs` file into a seperate module named `enums` (which is more appropriate compared with `senum`) with three files `country.rs`, `types.rs`, `misc.rs`, check [enums mod.rs file](./src/enums/mod.rs) for details.
```
marioortizmanero, line 69: Typo: "separate", not "seperate"

CHANGELOG.md
```diff
@@ -38,6 +40,33 @@ If we missed any change or there's something you'd like to discuss about this ve
...
+	- `current_user_saved_albums_add`
+	- `current_user_saved_albums_delete`
+	- `current_user_saved_albums_contains`
+	- `user_follow_artists`
+	- `user_unfollow_artists`
+	- `user_artist_check_follow`
+	- `user_follow_users`
+	- `user_unfollow_users`
+	- `audios_features`
+- Reexport `model` module to allow user to write `rspotify::model::FullAlbum` instead of  `rspotify::model::FullAlbum`.
```
marioortizmanero, line 68: I think you got confused here

CHANGELOG.md
```diff
@@ -38,6 +40,33 @@ If we missed any change or there's something you'd like to discuss about this ve
...
+	- `current_user_saved_tracks_add`
+	- `current_user_saved_albums_add`
+	- `current_user_saved_albums_delete`
+	- `current_user_saved_albums_contains`
+	- `user_follow_artists`
+	- `user_unfollow_artists`
+	- `user_artist_check_follow`
+	- `user_follow_users`
+	- `user_unfollow_users`
+	- `audios_features`
```
marioortizmanero, line 67: Now that I realize, `audios_features` should be renamed to `audio_features`. Should we do that in this PR or leave it for later?

examples/with_refresh_token.rs
```diff
@@ -43,7 +43,7 @@ async fn do_things(spotify: Spotify) {
     );
 
     spotify
-        .user_unfollow_artists(&artists)
+        .user_unfollow_artists(artists.clone())
```
marioortizmanero, line 46: This `clone` is not necessary.

src/client.rs
```diff
@@ -1309,10 +1339,13 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn user_artist_check_follow(&self, artsit_ids: &[String]) -> ClientResult<Vec<bool>> {
+    pub async fn user_artist_check_follow<'a>(
+        &self,
+        artsit_ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<Vec<bool>> {
         let url = format!(
             "me/following/contains?type=artist&ids={}",
-            artsit_ids.join(",")
+            artsit_ids.into_iter().collect::<Vec<&str>>().join(",")
```
marioortizmanero, line 1348: The `collect` can usually be simplified to `collect::<Vec<_>>()`.

src/client.rs
```diff
@@ -1325,8 +1358,14 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn user_follow_users(&self, user_ids: &[String]) -> ClientResult<()> {
-        let url = format!("me/following?type=user&ids={}", user_ids.join(","));
+    pub async fn user_follow_users<'a>(
+        &self,
+        user_ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<()> {
+        let url = format!(
+            "me/following?type=user&ids={}",
+            user_ids.into_iter().collect::<Vec<&str>>().join(",")
```
marioortizmanero, line 1367: Same here, the collect can be simplified.

src/client.rs
```diff
@@ -1339,8 +1378,14 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn user_unfollow_users(&self, user_ids: &[String]) -> ClientResult<()> {
-        let url = format!("me/following?type=user&ids={}", user_ids.join(","));
+    pub async fn user_unfollow_users<'a>(
+        &self,
+        user_ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<()> {
+        let url = format!(
+            "me/following?type=user&ids={}",
+            user_ids.into_iter().collect::<Vec<&str>>().join(",")
```
marioortizmanero, line 1387: The collect can be simplified.

src/client.rs
```diff
@@ -1894,8 +1939,8 @@ impl Spotify {
     ///
     /// [Reference](https://developer.spotify.com/console/put-current-user-saved-shows)
     #[maybe_async]
-    pub async fn save_shows(&self, ids: Vec<String>) -> ClientResult<()> {
-        let joined_ids = ids.join(",");
+    pub async fn save_shows<'a>(&self, ids: impl IntoIterator<Item = &'a str>) -> ClientResult<()> {
+        let joined_ids = ids.into_iter().collect::<Vec<&str>>().join(",");
```
marioortizmanero, line 1943: The collect can be simplified.

src/enums/types.rs
```diff
@@ -0,0 +1,254 @@
...
+use std::error;
+use std::fmt;
+use std::str::FromStr;
+
+#[derive(Clone, Debug)]
+pub struct Error {
+    kind: ErrorKind,
+}
+
+impl Error {
```
marioortizmanero, line 11: Couldn't this Error be removed and merge the ErrorKind with ClientError? Something like `ParseEnum` would be fine IMO.

src/client.rs
```diff
@@ -1952,14 +1997,17 @@ impl Spotify {
...
-        ids: Vec<String>,
+        ids: impl IntoIterator<Item = &'a str>,
         market: Option<Country>,
     ) -> ClientResult<SeversalSimplifiedShows> {
         // TODO: This can probably be better
         let mut params = Query::with_capacity(1);
-        params.insert("ids".to_owned(), ids.join(","));
+        params.insert(
+            "ids".to_owned(),
+            ids.into_iter().collect::<Vec<&str>>().join(","),
```
marioortizmanero, line 2009: The collect can be simplified.

src/client.rs
```diff
@@ -2031,13 +2079,16 @@ impl Spotify {
...
         &self,
-        ids: Vec<String>,
+        ids: impl IntoIterator<Item = &'a str>,
         market: Option<Country>,
     ) -> ClientResult<SeveralEpisodes> {
         let mut params = Query::with_capacity(1);
-        params.insert("ids".to_owned(), ids.join(","));
+        params.insert(
+            "ids".to_owned(),
+            ids.into_iter().collect::<Vec<&str>>().join(","),
```
marioortizmanero, line 2090: The collect can be simplified.

src/client.rs
```diff
@@ -2052,9 +2103,15 @@ impl Spotify {
...
-    pub async fn check_users_saved_shows(&self, ids: Vec<String>) -> ClientResult<Vec<bool>> {
+    pub async fn check_users_saved_shows<'a>(
+        &self,
+        ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<Vec<bool>> {
         let mut params = Query::with_capacity(1);
-        params.insert("ids".to_owned(), ids.join(","));
+        params.insert(
+            "ids".to_owned(),
+            ids.into_iter().collect::<Vec<&str>>().join(","),
```
marioortizmanero, line 2113: The collect can be simplified.

src/client.rs
```diff
@@ -2068,12 +2125,12 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn remove_users_saved_shows(
+    pub async fn remove_users_saved_shows<'a>(
         &self,
-        ids: Vec<String>,
+        ids: impl IntoIterator<Item = &'a str>,
         market: Option<Country>,
     ) -> ClientResult<()> {
-        let joined_ids = ids.join(",");
+        let joined_ids = ids.into_iter().collect::<Vec<&str>>().join(",");
```
marioortizmanero, line 2133: The collect can be simplified.

src/client.rs
```diff
@@ -1294,8 +1318,14 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn user_unfollow_artists(&self, artist_ids: &[String]) -> ClientResult<()> {
-        let url = format!("me/following?type=artist&ids={}", artist_ids.join(","));
+    pub async fn user_unfollow_artists<'a>(
+        &self,
+        artist_ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<()> {
+        let url = format!(
+            "me/following?type=artist&ids={}",
+            artist_ids.into_iter().collect::<Vec<&str>>().join(",")
```
marioortizmanero, line 1327: This collect can be simplified

src/client.rs
```diff
@@ -1280,8 +1298,14 @@ impl Spotify {
...
     #[maybe_async]
-    pub async fn user_follow_artists(&self, artist_ids: &[String]) -> ClientResult<()> {
-        let url = format!("me/following?type=artist&ids={}", artist_ids.join(","));
+    pub async fn user_follow_artists<'a>(
+        &self,
+        artist_ids: impl IntoIterator<Item = &'a str>,
+    ) -> ClientResult<()> {
+        let url = format!(
+            "me/following?type=artist&ids={}",
+            artist_ids.into_iter().collect::<Vec<&str>>().join(",")
```
marioortizmanero, line 1307: This collect can be simplified

marioortizmanero (commented):
I've left some comments about problems I found when looking at the code. There's two things I want to point out:

* Why is `enum` a separate module? Can't this be included in the `model` module as well, as after all, they are part of the models? Or am I just wrong here?
* The conversions to/from enums are incredibly repetitive. Can we use a custom macro or something like [`strum`](https://docs.rs/strum/0.19.5/strum/) instead?

CHANGELOG.md
ramsayleung: I could fix it in this PR.

CHANGELOG.md
ramsayleung: Sorry for not getting your point.

CHANGELOG.md
marioortizmanero: sure, that's probably easier

CHANGELOG.md
marioortizmanero: I just don't understand this line. Both import paths are the same.

examples/with_refresh_token.rs
```diff
@@ -43,7 +43,7 @@ async fn do_things(spotify: Spotify) {
     );
 
     spotify
-        .user_unfollow_artists(artists.clone())
+        .user_unfollow_artists(artists)
```
marioortizmanero: Ah my bad. It seems I was wrong. You do need the `clone`

CHANGELOG.md
ramsayleung: Ooops, I get it now. I'll fix it.

ramsayleung:
>  Why is enum a separate module? Can't this be included in the model module as well, as after all, they are part of the models? Or am I just wrong here?

Actually, the `enum` is a single file name `senum.rs`(means `enums`, a bad joke), but with more and more `enum` added, the `senum` file grown up to be a huge one, but I separated it into a module, with multiple 

> The conversions to/from enums are incredibly repetitive. Can we use a custom macro or something like strum instead?

Of course we could, I could take time to figure out how to create a decent macro.

marioortizmanero:
> Actually, the enum was a single file named senum.rs(which means enums, a bad joke) before, but with more and more enums added, the senum file grown up to be a huge one, but I separated it into a module, with multiple files.

But why is it not merged with the `model` module? It kinda seems like the same to me, but I'm not sure if I understand it correctly.

ramsayleung:
> But why is it not merged with the model module? It kinda seems like the same to me, but I'm not sure if I understand it correctly.

Yes, we could, perhaps it's more consistent to move `enums` module into `model` module.

Another problem I want to discuss is that which import path do you prefer,  `rspotify::model::FullAlbum` or `rspotify::FullAlbum`. 

The former one is longger, and explicit, the later one is shorter and shorter. I personnally perfer to the former one.

marioortizmanero:
I think it's better to keep that in the `model` module as well.

src/enums/types.rs
```diff
@@ -0,0 +1,254 @@
...
+            "track" => Ok(SearchType::Track),
+            "playlist" => Ok(SearchType::Playlist),
+            "show" => Ok(SearchType::Show),
+            "episode" => Ok(SearchType::Episode),
+            _ => Err(Error::new(ErrorKind::NoEnum(s.to_owned()))),
+        }
+    }
+}
+
+#[test]
```
marioortizmanero, line 227: these test functions should also be inside a `mod test`.

src/client.rs
ramsayleung: Yep, it's a acceptable choice to use `use super::model::*`, or we have to import a lot of model files individually.

ramsayleung:
I have refactored `enums` module with `strum` crate, reducing line-of-code from 7396 to 6972, generated by `tokei`. It's much clear and readable than before, but we have introduced a new dependency, I think about that will it increase the compilation time ? The number of compilation units increased from 159 to 163 with default feature.

marioortizmanero:
That's great! Have you measured the compilation times, rather than the number of units?

ramsayleung:
Yes, I have measured the compilation times:

```sh
git checkout 72c6ee3 # the last commit without `strum` crate.
> hyperfine --warmup 2 'cargo clean && cargo build --no-default-features --features client-reqwest' -i
Benchmark #1: cargo clean && cargo build --no-default-features --features client-reqwest
  Time (mean ± σ):     86.715 s ±  3.165 s    [User: 276.518 s, System: 14.973 s]
  Range (min … max):   82.019 s … 91.968 s    10 runs

git checkout ramsay_clean_up_rspotify 
> hyperfine --warmup 2 'cargo clean && cargo build --no-default-features --features client-reqwest' -i
Benchmark #1: cargo clean && cargo build --no-default-features --features client-reqwest
  Time (mean ± σ):     88.646 s ±  1.688 s    [User: 285.274 s, System: 15.676 s]
  Range (min … max):   84.932 s … 90.441 s    10 runs
```

CHANGELOG.md
ramsayleung: Well, it turns out that it's not a problem, we have both `audio_features` and `audios_features` endpoint:

+ audio_features: Get audio features for a track
+ audios_features: Get Audio Features for Several Tracks

marioortizmanero: Ah ok. How about `track` instead of `audio`? "audios" seems off to me, perhaps `track_features` and `tracks_features` would be more fitting

src/model/enums/misc.rs
```diff
@@ -0,0 +1,105 @@
+use serde::{Deserialize, Serialize};
+use strum::{AsRefStr, Display, EnumString};
+
+/// disallow: interrupting_playback, pausing, resuming, seeking, skipping_next, skipping_prev, toggling_repeat_context, toggling_shuffle, toggling_repeat_track, transferring_playback
+#[derive(
+    Clone, Serialize, Deserialize, Copy, PartialEq, Eq, Debug, Hash, EnumString, AsRefStr, Display,
```
marioortizmanero, line 6: May I ask why we use `AsRefStr` instead of `ToString` for these derive macros? What is the `disallow: ...` part in the comment as well?

src/model/enums/misc.rs
```diff
@@ -0,0 +1,105 @@
+use serde::{Deserialize, Serialize};
+use strum::{AsRefStr, Display, EnumString};
+
+/// disallow: interrupting_playback, pausing, resuming, seeking, skipping_next, skipping_prev, toggling_repeat_context, toggling_shuffle, toggling_repeat_track, transferring_playback
```
marioortizmanero, line 4: What is the `disallow: ...` part in the comment?

marioortizmanero, line 4: What is the `disallow: ...` part in the comment?

src/model/enums/misc.rs
marioortizmanero, line 6: May I ask why we use `AsRefStr` instead of `ToString` for these derive macros?

src/model/enums/country.rs
```diff
@@ -0,0 +1,525 @@
+use serde::{Deserialize, Serialize};
+use strum::{AsRefStr, Display, EnumString};
+
+///ISO 3166-1 alpha-2 country code, [wiki about ISO 3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
+///Source from [country-list](https://datahub.io/core/country-list)
+#[derive(
+    Clone, Serialize, Deserialize, Copy, PartialEq, Eq, Debug, EnumString, AsRefStr, Display,
+)]
+pub enum Country {
```
marioortizmanero, line 9: Isn't there a crate for this already? I would really prefer that if possible. It saves us 500 lines of untested and unmaintained code, compared to a well known crate. `aspotify` seems to use `isocode`: https://docs.rs/aspotify/0.6.1/aspotify/enum.CountryCode.html

src/model/enums/types.rs
```diff
@@ -0,0 +1,155 @@
...
+/// currently_playing_type: track, episode, ad, unknown.
+#[derive(
+    Clone, Serialize, Deserialize, Copy, PartialEq, Eq, Debug, EnumString, AsRefStr, Display,
+)]
+#[serde(rename_all = "snake_case")]
+#[strum(serialize_all = "snake_case")]
+pub enum CurrentlyPlayingType {
+    Track,
+    Episode,
+    #[strum(serialize = "ad")]
```
marioortizmanero, line 53: Why is this done with `strum` and not `serde`?

src/model/enums/mod.rs
```diff
@@ -0,0 +1,18 @@
+//! All Enums for Rspotify
+pub mod country;
+pub mod misc;
+pub mod types;
+
+pub use country::Country;
+
+pub use misc::DisallowKey;
```
marioortizmanero, line 8: Can these be grouped up with `use thing::{a,b,c}`?

src/model/enums/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+pub use misc::IncludeExternal;
+pub use misc::RepeatState;
+pub use misc::TimeRange;
+
+pub use types::AdditionalType;
```
marioortizmanero, line 13: Same as above

src/model/enums/misc.rs
ramsayleung: ｀disallow｀ just means `disallow key:`, I would rename it to `disallow key`.

src/model/enums/misc.rs
ramsayleung: Actually, I can't tell the difference between `EnumString` and `Tostring`, which are almost same, so I choose `EnumString` with additional attribute `#[strum(serialize_all = "snake_case")]`.

`AsRefStr` is another trait to derive, which performs as same as the old function `as_str()`. Thus, instead of prefer `Tostring` to `AsRefStr`, I prefer to keep them both, but I choose `EnumString` instead of `ToString`.

src/model/enums/country.rs
ramsayleung: I don't think it's necessary to add a new crate just to take one `enum` from this crate, which will break the consistency of `enums` module, some enums are defined in `enums` module, and others are defined in another crate, adding such a new crate doesn't pay off.

src/model/enums/types.rs
ramsayleung: Yes, it's a problem, I'll fix it.

src/model/enums/mod.rs
ramsayleung: Sure!

src/model/enums/misc.rs
marioortizmanero: I still don't quite understand it, sorry. What's that for?

ramsayleung: Disallow means the [disallows object](https://developer.spotify.com/documentation/web-api/reference/object-model/#disallows-object), which is used in the endpoint [`get-users-currently-playing-track`](https://developer.spotify.com/documentation/web-api/reference/player/get-the-users-currently-playing-track/), is is clear to you now?

CHANGELOG.md
ramsayleung: Yep, it's a more suitable name, I will rename `audio_features`, `audios_features` and `audio_analysis` to `track_features`, `tracks_features` and `track_analysis`.

src/model/enums/misc.rs
marioortizmanero: Ahhh okay, yes, sorry. The link to the reference helped. For some reason I thought it was something for the linter.

Can you add the references for these structs/enums like the endpoints? They are usually like this:

```rust
/// Endpoint: description
///
/// ...
///
/// [Reference](reference)
```

just for consistency

src/model/enums/misc.rs
marioortizmanero: The difference is that `EnumString` implements `FromStr`, and `ToString` implents `ToString`. So you can't just choose between these two because they do different things. If anything, the choice is between `ToString`, which returns a `String`, and `AsRefStr`, which returns a `&str`. But IMO the former makes more sense in the context of what we need: to convert something into a string.

But that makes me wonder what these derives are even for. `aspotify` for example doesn't seem to implement `ToString` and `FromStr`. Are they needed for serialization or something? Isn't that done by `serde`?

src/model/enums/misc.rs
ramsayleung: Yep, I have already updated this. This comments is outdated.

src/model/enums/misc.rs
ramsayleung: >  Isn't that done by serde?

No, it's nothing related to `serialization`, `to_string()` is for setting something like a query parameter:
```rust
    #[maybe_async]
    pub async fn tracks<'a>(
        &self,
        track_ids: impl IntoIterator<Item = &'a str>,
        market: Option<Country>,
    ) -> ClientResult<FullTracks> {
        // TODO: this can be improved
        let mut ids: Vec<String> = vec![];
        for track_id in track_ids {
            ids.push(self.get_id(Type::Track, track_id));
        }

        let mut params = Query::new();
        if let Some(market) = market {
            params.insert("market".to_owned(), market.to_string());
        }

        let url = format!("tracks/?ids={}", ids.join(","));
        let result = self.get(&url, None, &params).await?;
        self.convert_result(&result)
    }

```
`enums` is not only for model, but also used in somewhere else.

marioortizmanero: > No, it's nothing related to serialization, to_string() is for setting something like a query parameter:

OK, thanks for clearing it up. So wouldn't it make more sense to convert to a string with `.to_string()` in that case instead of `.as_ref()`? AFAIK the latter is mostly used in smart pointers like `Box`.

src/model/enums/misc.rs
marioortizmanero: Great. Can you make it 80 characters long as always?

ramsayleung: Yes, of course. I think it's a better way to let `rustfmt` handle it :)

src/model/enums/misc.rs
ramsayleung: > AFAIK the latter is mostly used in smart pointers like Box. Some of your enums derive both AsRefStr and Display; I think we would be fine with just a single ToString.

Yep, good point, I would replace all `EnumString, AsRefStr, Display` to `ToString`, your point makes sense :)

marioortizmanero: I've been investigating and in fact, we could avoid `strum` altogether and use this from `serde_json` - https://docs.serde.rs/serde_json/ser/fn.to_string.html

marioortizmanero: Great! That should reduce compile times as well.

src/model/enums/misc.rs
marioortizmanero: rustfmt doesn't modify comments AFAIK

src/model/enums/misc.rs
marioortizmanero: All I've found is this crate - https://docs.rs/serde_plain/0.3.0/serde_plain/fn.to_string.html - but I'm not sure if it's better than using `strum`. It saves us from repeating the lowercase conversions and renames, and it seems to be lighter than `strum`, but you'd have to use `serde_plain::to_string(&val)?` instead of `var.to_string()`.

ramsayleung: IMHO, this syntax is a little bit weird, perhaps we could find another way to do so.
 
> you'd have to use serde_plain::to_string(&val)? instead of var.to_string().

marioortizmanero: I think it's impossible:

1. `Serialize` is an external trait from `serde`
2. `ToString` is an external trait from `std`

Thus, a crate like `serde_json` can't have a blanket implementation of `ToString` for all `T: Serialize` types because both are external to it. If anything, it could create an equivalent trait for `ToString`, like `OurToString`, but that wouldn't play well with generic types over `ToString`.

ramsayleung: Yeah, I agree with you, `strum` is a solid solution for our case. By the way, I have pushed some new commits, including a bit model modification. I will keep refactoring model after this PR merged.

marioortizmanero (approved):
I think this is ready to be merged.

ramsayleung:
Merged :) 

I think I will take times to cross off more items in #127 :)

PR #133, opened 2020-10-08
Fix playlist related endpoints

Sydpy:
Most of the API for handling playlists required a user ID in addition to the playlist ID. However, the Spotify API only requires the playlist ID for its endpoints, [see here](https://developer.spotify.com/documentation/web-api/reference/playlists/).

I suppose that in a previous version of the Spotify API, the owner ID was needed to correctly identify a playlist, but it seems to no longer be the case.

This PR fixes the calls to the playlist related actions, and renames the methods of the client for more clarity.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -896,21 +896,16 @@ impl Spotify {
     /// Add the current authenticated user as a follower of a playlist.
     ///
     /// Parameters:
-    /// - playlist_owner_id - the user id of the playlist owner
     /// - playlist_id - the id of the playlist
     ///
     /// [Reference](https://developer.spotify.com/web-api/follow-playlist/)
     #[maybe_async]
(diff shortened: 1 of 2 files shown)


ramsayleung:
Thanks for your contribution, and I think this PR needs to wait for #129 , which refactors core codebase of `rspotify`. it will cause some conflicts if merge this PR first :)

marioortizmanero:
Exactly. I'm currently a bit busy with the degree so I haven't worked on it much lately. We accept any kind of contributions in there as well ;)

Sydpy:
Thanks for the quick responses :), #129 looks like a big PR, I think I'll have a hard time getting into it since I'm new to this crate but I'll try to take a look at it

Sydpy:
I rebased this branch on master to take into account the recent changes

ramsayleung:
Merged, Cheers :)

Issue #150, opened 2020-11-08
Authorization Code Flow with Proof Key for Code Exchange (PKCE) is missing


martinmake:
It is a very exciting *Authorization Flow* for developers of server-less applications that use *Spotify API* directly.

Not only will the users of these applications no longer have to setup their own *Spotify Applications*,
but also the creators will get to see usage metrics of their applications in their *Spotify Dashboard*.

You can find out how it works [here](https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-code-flow-with-proof-key-for-code-exchange-pkce) and [here](https://docs.rs/oauth2/3.0.0/oauth2/#getting-started-authorization-code-grant-w-pkce).

ramsayleung:
Yes, `rspotify` doesn't support PKCE now, and it seems you are going to contribute it to `rspotify` #151 . Before you start to work, would you like to illustrate your though firstly? just like what @marioortizmanero did before, it's his proposal:

+ https://github.com/ramsayleung/rspotify/issues/112
+ https://github.com/ramsayleung/rspotify/pull/120
+ https://github.com/ramsayleung/rspotify/pull/129

The template of your proposal could be:

+ What are you going to build ?
+ Why would you like to build it?
+ How to implement your proposal?

As for how to implement your idea, now there are different ways to approach your idea, just to brainstorm the working solutions as much as possible, and try to clarify their pros and cons, then choose the best one(from where you stand). for example:

+ Solution A: 
  - pros: xx
  - cons: xx
+ Solution B: xxx
+ Solution C: xxx

You could take a look at this issue #134, it makes a great example.

And then inviting other developers to review your proposal :)

PR #151, opened 2020-11-08
Adding Authorization Code Flow with Proof Key for Code Exchange (PKCE)

martinmake:
I would start with a discussion about the user interface.
I've added an example for this feature so we can have the same context.

What do you think about it?

Should we use:
* `pub enum AuthorizationFlow` to choose between *Authorization Flows*, or
* builders/functions, or
* should  it be inferred from `secret: Option<String>` (won't scale well with other *Authorization Flows*)?


marioortizmanero:
Hi, sorry for not replying earlier. I wanted to read about the PKCE authentication process beforehand. Maybe we should re-think the authentication process? I didn't really like how it turned out in the rewrite because we need stuff like `InvalidAuth` to for example make sure the `me` endpoint isn't accessed by a client with the Client Credentials  Flow. I was thinking that we could take advantage of Rust's features and only implement these endpoints for some flows and make to make it more type-safe with a different architecture.

I'll open a new issue soon with some ideas where we can discuss how to make it work.

marioortizmanero:
This is being discussed at #173 now. Let me know what you think!

marioortizmanero:
Closing this, as #173 is more or less accepted.

marioortizmanero:
This shouldn't have been closed because PKCE was left as a TODO In that PR. Reopening.

marioortizmanero:
Ah sorry! I thought I was in the issue #150 but this was the old PR. Nevermind.

Issue #160, opened 2020-11-24
There are a lot of unnecessary allocations in the library


kstep:
**Is your feature request related to a problem? Please describe.**
There are a lot of places with unnecessary clones/vector constructions/other allocations.
E.g. `get_id()` function does things like this:
```
if _type.to_string() != fields[len - 2] { ... }
```
This leads to unnecessary allocation of a piece of memory, immediately thrown away.
It probably should be `if _type.as_str() != fields[len -2]`.

Also, there are a lot of places with `something_ids.into_iter().collect::<Vec<_>>().join(",")`, which allocates a vector just to throw it away. It can be replaced with just `something_ids.into_iter().join(",")` using `itertools` crate, which is very efficient, trying hard to allocate as little memory as possible and joins elements in one pass instead of two.

Another example is `&format!("me/player/repeat?state={}", state.to_string())`, why allocate a transient string here? It should be something like `&format!("me/player/repeat?state={}", state.as_str())`.

**Describe the solution you'd like**
Use `&'static str` to get a string representation of enums. Use slices and iterators without transient allocations with `collect()` wherever possible. Use enums to compare things instead of strings. Avoid `.to_string()` unless totally necessary.

**Describe alternatives you've considered**
Leave things as they are. But I think using unnecessary allocations in languages like Rust is a sin, as doing so is losing the point of such a low-level language at all.

marioortizmanero:
I completely agree, I would like to reduce the number of allocations in this crate as well. But IMO that's currently very low priority, we're under some refactoring right now and lots of stuff could change. For instance I wanted to rewrite `get_id` completely because the function itself is quite messy. Maybe after we release the next version we can focus on things like this.

For simple things you can always open a PR and I'll gladly review it, but I wouldn't want to lose too much time on this yet.

marioortizmanero:
This is now too old to be relevant. If anyone still thinks there are too many allocations (only if removing them doesn't hurt usability), please open a new issue with examples from the codebase. Thanks!

PR #161, opened 2020-11-29
Initial id type proposal

kstep:
## Description

This is a refactoring/change proposal. Introduces several changes and refactorings to the API:

- new `Id` type to represent Spotify id/URI,
- a new `ClientError` variant: `InvalidId(IdError)`,
- removed `.get_id()` and `.get_uri()` methods,
- introduced new dependency: `itertools`, used it in API methods,
- added `AsRefStr` for `Type`.

The main point of the change is to introduce `Id` type. This type abstracts out the following Spotify id/URI operations:

- Spotify id/URI parsing and validation,
- Spotify URI formatting,
- zero-copy: keeps just a slice of original data,
- throws an `IdError` error if id/URI is invalid.

This type replaces both `.get_id()` and `.get_uri()` methods.

Also, `Spotify` methods are changed to use this type instead of `.get_id()` and `.get_uri()`.
This type, being zero-copy, should also reduce the number of allocations.

This change is a proposal only!

## Open Questions/TODO

- I'm not sure where should I put the `Id` and `IdError` types, put them into `model/mod.rs` for now.
- I'm not sure about tests for `Id` placement, so I left them in `model/mod.rs` for now.
- Maybe more test-cases should be added.
- Maybe it's better to use `Cow<'_, str>` and `Into<Cow<'_, str>>` types for internal `Id` data representation and method arguments.

## Motivation and Context

As per TODO comments in code, and as a part of the refactoring effort as per #127 and #160, I decided to propose a new type to abstract out Spotify id/URI operations. Overall improvements:

- better input data validation (ids, URIs): invalid ids and URIs now produce errors,
- better type-safe Spotify ids representation,
- fewer allocations during Spotify ids parsing and processing.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

- I updated tests related to `.get_id()`/`.get_uri()` methods.
- I ran all the enabled tests using default configuration (with my own credentials in `.env`) with `cargo test` command.

---
This change is [Reviewable](https://reviewable.io/reviews/ramsayleung/rspotify/161)


src/client.rs
```diff
@@ -15,6 +15,37 @@ use super::http::{BaseClient, Query};
...
+    where
+        Self: Iterator<Item = T>,
+        F: Fn(T) -> Result<R, E>,
+        R: AsRef<str>,
+    {
+        self.map(func).try_join(sep)
+    }
+}
+
+impl<T> TryJoin for T where T: Itertools {}
```
ramsayleung, line 48: I am a little bit confused about the `TryJoin` trait and `may_try_join` function, what kind of problem will it solve ?

As for `itertools` crate, is it possible to replace it with a customized trait instead of introducing a new crate? Since we are working on reducing the compile time of `rspotify` :)

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
     Episode(show::FullEpisode),
 }
 
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct Id<'id> {
```
ramsayleung, line 54: I am thinking about that is `Id` a good name to describe its functionality? and it's same to `from_id_or_uri`

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+
+    pub fn id(&self) -> &str {
+        &self.id
+    }
+
+    pub fn uri(&self) -> String {
+        format!("spotify:{}:{}", self._type.as_ref(), self.id)
+    }
+
+    pub fn from_id_or_uri<'a, 'b: 'a>(_type: Type, id_or_uri: &'b str) -> Result<Id<'a>, IdError> {
```
ramsayleung, line 96: IMHO, I think `from_id_or_uri` should be more explicit, just use this function to `get_id` or `get_uri`, instead of trying them both, it's a bit implicit.

src/client.rs
kstep: It reduces code duplication, it just repeats all the way down in half of the API methods.

kstep: As for `itertools` dependency, I'll try to remove it.

kstep: I rewrote this trait into a single method.

src/model/mod.rs
kstep: The `Id` type represents spotify object id, so yes, I think it's the most correct name for it.
Other possible names are `SpotifyId` (it's a little bit verbose, as we are already in a crate named `rspotify`), `Uri`, or `SpotifyUri` (too verbose, as well as `SpotifyId`).

src/model/mod.rs
kstep: One of the problems it solves, it unifies URI and id arguments in methods, so a user can use either Spotify id or URI, no matter what real Spotify API endpoint requires. So this method accepts either id or URI from one of the API endpoint methods, hence the name.

src/model/mod.rs
kstep: Do you have better proposals?

marioortizmanero: I don't see how `Id` isn't an appropiate name either. `SpotifyId` is redundant.

src/model/mod.rs
marioortizmanero: I think it's fine to have `from_id_or_uri` as long as we also have `from_id` and `from_uri`, just so that the user can call it as a convenience method for whenever they don't know what you're parsing (say, user input).

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+    }
+}
+
+impl AsRef<str> for Id<'_> {
+    fn as_ref(&self) -> &str {
+        self.id
+    }
+}
+
+impl Id<'_> {
```
marioortizmanero, line 83: This type should have more documentation regarding what's considered an ID and what's an URI. It should also note that URIs may use `/` or `:` as separators, and that IDs are made up of alphanumeric characters only.

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+        match Self::from_uri(id_or_uri) {
+            Ok(id) if id._type == _type => Ok(id),
+            Ok(_) => Err(IdError::InvalidType),
+            Err(IdError::InvalidPrefix) => Self::from_id(_type, id_or_uri),
+            Err(error) => Err(error),
+        }
+    }
+
+    pub fn from_id<'a, 'b: 'a>(_type: Type, id: &'b str) -> Result<Id<'a>, IdError> {
+        if id.chars().all(|ch| ch.is_ascii_alphanumeric()) {
```
marioortizmanero, line 106: I think you can use `all(char::is_ascii_alphanumeric)` directly here.

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+            Ok(Id { _type, id })
+        } else {
+            Err(IdError::InvalidId)
+        }
+    }
+
+    pub fn from_uri<'a, 'b: 'a>(uri: &'b str) -> Result<Id<'a>, IdError> {
+        if let Some((tpe, id)) = if let Some(uri) = uri.strip_prefix("spotify:") {
```
marioortizmanero, line 114: Having nested `if let`s makes the code much harder to read. Is it possible to break this up a bit? For instance we could:

1. Strip the starting `spotify`
2. Compare the first character. If it's not `:` or `/`, return an error.
3. Use that character to split the remaining slice into what should be:
    1. An empty slice
    2. The type, which can be parsed
    3. The ID, which can be passed to `from_id`

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+            uri.rfind(':').map(|mid| uri.split_at(mid))
+        } else if let Some(uri) = uri.strip_prefix("spotify/") {
+            uri.rfind('/').map(|mid| uri.split_at(mid))
+        } else {
+            return Err(IdError::InvalidPrefix);
+        } {
+            let _type = match tpe {
```
marioortizmanero, line 121: This should be a `FromStr` implementation for `Type`, right?

src/model/mod.rs
```diff
@@ -48,8 +50,145 @@ pub enum PlayingItem {
...
+pub enum IdError {
+    InvalidPrefix,
+    InvalidFormat,
+    InvalidType,
+    InvalidId,
+}
+
+impl std::fmt::Display for Id<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.write_str("spotify:")?;
```
marioortizmanero, line 69: You could just use `write_fmt` here and undo deriving `AsRefStr` for `Type`.

src/client.rs
```diff
@@ -16,6 +16,36 @@ use super::json_insert;
 use super::model::*;
 use super::oauth2::{Credentials, OAuth, Token};
 
+pub trait TryJoin: Iterator {
```
marioortizmanero, line 19: I still don't get why this is necessary at all. Isn't `Option.map(|x| x.join())` equivalent?

src/client.rs
```diff
@@ -202,8 +195,8 @@ impl Spotify {
     /// [Reference](https://developer.spotify.com/web-api/get-track/)
     #[maybe_async]
     pub async fn track(&self, track_id: &str) -> ClientResult<FullTrack> {
-        let trid = self.get_id(Type::Track, track_id);
-        let url = format!("tracks/{}", trid);
+        let trid = Id::from_id_or_uri(Type::Track, track_id)?;
```
marioortizmanero, line 198: Instead of this, the function should take an `Id` parameter, and we can later check that it's a `Type::Track`. That way, the user may initialize `Id` however they want. For example, if they are sure it's an URI there's no need to use `from_id_or_uri`.

src/model/mod.rs
kstep: I don't like having an empty slice. It feels like an unnecessary splitting job.

src/model/mod.rs
kstep: If I don't use `AsRefStr`, I'd have to use either `Display` or `ToString`. I have a hard "no" for `ToString`, as it allocates, and the whole point of this PR is to avoid unnecessary allocations. So if I remove `AsRefStr` I will end up implementing `Display` on `Type` anyway. Do you think it's better?

src/model/mod.rs
kstep: I see. So you want to expose the `Id` type to the user and make it a part of public API? This first proposal was intended as part of an internal API. If you think it's better to make it public, then I agree, we need:

1. Good documentation on `Id` type
2. Make all methods accept `Id` instead of `&str` when Spotify id or URI is required

src/model/mod.rs
kstep: No, I can't. `all()` requires `fn(&char) -> bool`, and `char::is_ascii_alphanumeric()` is `fn(char) -> bool`.

src/model/mod.rs
kstep: By the way, since I already use `Display` instead of `ToString` for `Type`, maybe we should go to the end and use it instead of `ToString` for all other enums?

src/client.rs
kstep: No, it's not. To do `something.join(",")`, `something` must be an allocated `Vec<impl Display>`. This implementation is more efficient, it avoids `Vec<_>` allocation.

kstep:
Is it a good idea to move `Id` parsing/validation to our end users? If we accept `Id` everywhere, user's will have to do some work to parse ids and create `Id` instances. It makes our library less user-friendly. Are you sure it's OK?

src/model/mod.rs
marioortizmanero: Yes, sorry if I didn't make my idea clearer before. Having `Id` as public would give the user more options on these kinds of parameters, and makes use of the type safety Rust provides.

Why not use `String` rather than `Cow<&str>`? That's how these structs usually go.

Very thorough documentation, by the way! I love it.

marioortizmanero: Ah okay I just read in another comment that what you want with this is to avoid allocations. I'm not that experienced with that and I'm not sure at what point it's worth just not using `String` at all. I'd have to read more about it.

src/model/mod.rs
marioortizmanero: Ah okay, I didn't realize the `&` in the type signature. Don't worry about it then.

marioortizmanero:
> Is it a good idea to move Id parsing/validation to our end users? If we accept Id everywhere, user's will have to do some work to parse ids and create Id instances. It makes our library less user-friendly. Are you sure it's OK?

IMO passing `Id` to the endpoints instead of `&str` is beneficial because the user can choose how to parse it when they know whether it's an ID or a URI. It's also a bit more type-safe.

I would love to hear @ramsayleung's opinion about this.

marioortizmanero:
I can see that you're really avoiding allocations to progress towards #160. I wanted to disscuss about this before we make any changes because it's a somewhat complicated topic in Rust. I've been trying to learn about when to use `&str` and when not to and I've come across a few sources:

* https://www.youtube.com/watch?v=gpyl_nEomDQ this video measures the performance differences between using `String` and `&str`. The end result is that by using `&str` you save around 50 nanoseconds in these benchmarks.
* https://github.com/clap-rs/clap/issues/1041 discusses this exact issue in the well-known `clap-rs` library. Unfortunately, there's not much discussion about it yet, but it's a good idea to learn more about what other libraries do, and their approaches.

I personally think that these kind of optimizations can be a pain for both programmers and users, and most times they aren't even noticeable. Specially if we end up propagating the use of `&str` or `Cow<&str>` to the model structs. I would like to see a benchmark for this library comparing these changes (not necessarily by you). Then, we can discuss whether they're worth it or not. As to your `Cow<&str>` recommendation, I would like to consider that as well, but not having seen it in any big libraries I would think it's not a good general-purpose choice. It's also mentioned in the `clap-rs` issue, but they didn't exactly succeed with it.

As far as I know, this library never intended to be a *performant API wrapper* for Spotify. In fact, we don't say anything about what the goals of Rspotify are in the README. Being the most popular wrapper in Rust for the Spotify API, my assumption would be that Rspotify aims to be the most flexible and feature-full one, which isn't fully compatible with being the most performant one.

kstep:
I see your point here. Actually, I gave `Cow<&str>` a try before, but the end result was very complex, so for the first version I ended up with a wrapper around `&str`. Anyway, if it's not too difficult to have fewer allocations, why not do it anyway?

marioortizmanero:
I've been thinking about this and I really don't like having to check what type is associated to an Id manually. Not only it's a runtime cost, we also have to implement a `TryJoin` iterator for better performance. So, what do you think about this:

```rust
use std::marker::PhantomData;

#[derive(Debug)]
pub struct Album {}
#[derive(Debug)]
pub struct Artist {}
#[derive(Debug)]
pub struct Track {}

#[derive(Debug)]
pub struct Id<'a, Ty> {
    pub id: &'a str,
    phantom: PhantomData<Ty>
}

impl<'a, T> Id<'a, T> {
    pub fn new(id: &'a str) -> Self {
        Id {
            id,
            phantom: PhantomData,
        }
    }
}

pub fn endpoint(id: Id<Album>) {
    println!("Id: {:?}", id);
}

fn main() {
    let id = Id::<Album>::new("asdf");
    endpoint(id);
}
```

Not sure if `PhantomData` is the appropiate type there, but if we make `Id` generic over the type it's related to then we don't have to check that the types are valid at all, it's all done at compile time :)

kstep:
Ok, I revamped `Id` type and added void types to represent ids (`idtypes` module).
Along the way I fixed `playlist_remove_specific_occurences_of_track()` method as part of an effort to make things more type-safe.

kstep:
The next step would be to replace all uri/id fields in models with id types :-)

marioortizmanero:
As always I'd rather perform some benchmarks before including an optimization
such as `Join`. Here's a simple benchmark using your custom `Join`, `std`
(collect + join), `itertools` and `fold`. All of them have a benchmark without
processing (as in, all that's needed is a `join`), and another with some (just a
`map`):

```rust
//! You'll need nightly for `#[bench]`.
#![feature(test)]

extern crate test;

pub trait Join: Iterator {
    fn join(&mut self, sep: &str) -> String
    where
        Self: Iterator,
        Self::Item: AsRef<str>,
    {
        if let Some(item) = self.next() {
            let value = item.as_ref();
            let (size, _) = self.size_hint();
            let cap = size * (sep.len() + value.len());

            let mut output = String::with_capacity(cap);
            output.push_str(value);

            for item in self {
                output.push_str(sep);
                output.push_str(item.as_ref());
            }

            output
        } else {
            String::new()
        }
    }
}

impl<T> Join for T where T: Iterator {}

#[cfg(test)]
mod tests {
    use test::Bencher;

    const MAX: u32 = 100_000;

    fn initial_vec() -> Vec<&'static str> {
        let mut v = Vec::new();
        for i in 1..=MAX {
            if i % 2 == 0 {
                v.push("even number here");
            } else {
                v.push("odd number over here");
            }
        }

        v
    }

    #[bench]
    fn bench_itertools(b: &mut Bencher) {
        use itertools::Itertools;
        let iter = initial_vec();

        b.iter(|| {
            iter.clone().into_iter().join(", ");
        });
    }

    #[bench]
    fn bench_itertools_processing(b: &mut Bencher) {
        use itertools::Itertools;
        let iter = initial_vec();

        b.iter(|| {
            iter.clone()
                .into_iter()
                .map(|s| format!("some processing: {}", s))
                .join(", ");
        });
    }

    #[bench]
    fn bench_fold(b: &mut Bencher) {
        let iter = initial_vec();

        b.iter(|| {
            iter.clone()
                .into_iter()
                .fold(String::new(), |a, b| a + ", " + &b);
        });
    }

    #[bench]
    fn bench_fold_processing(b: &mut Bencher) {
        let iter = initial_vec();

        b.iter(|| {
            iter.clone()
                .into_iter()
                .map(|s| format!("some processing: {}", s))
                .fold(String::new(), |a, b| a + ", " + &b);
        });
    }

    #[bench]
    fn bench_std(b: &mut Bencher) {
        let iter = initial_vec();

        b.iter(|| {
            iter.clone().into_iter().collect::<Vec<&str>>().join(", ");
        });
    }

    #[bench]
    fn bench_std_processing(b: &mut Bencher) {
        let iter = initial_vec();

        b.iter(|| {
            iter.clone()
                .into_iter()
                .map(|s| format!("some processing: {}", s))
                .collect::<Vec<_>>()
                .join(", ");
        });
    }

    #[bench]
    fn bench_custom(b: &mut Bencher) {
        use super::Join;
        let iter = initial_vec();

        b.iter(|| {
            iter.clone().into_iter().join(", ");
        });
    }

    #[bench]
    fn bench_custom_processing(b: &mut Bencher) {
        use super::Join;
        let iter = initial_vec();

        b.iter(|| {
            iter.clone()
                .into_iter()
                .map(|s| format!("some processing: {}", s))
                .join(", ");
        });
    }
}

fn main() {
    println!("Hello, world!");
}
```

And here are the results:

```
$ cargo bench
   Compiling iter v0.1.0 (/home/mario/Downloads/iter)
    Finished bench [optimized] target(s) in 0.66s
     Running /home/mario/.cache/cargo/release/deps/iter-98ee62b69dfffcbd

running 8 tests
test tests::bench_custom               ... bench:     599,477 ns/iter (+/- 56,302)
test tests::bench_custom_processing    ... bench:   5,719,576 ns/iter (+/- 414,945)
test tests::bench_fold                 ... bench:   1,590,366 ns/iter (+/- 169,231)
test tests::bench_fold_processing      ... bench:   5,550,425 ns/iter (+/- 164,123)
test tests::bench_itertools            ... bench:   2,251,977 ns/iter (+/- 171,879)
test tests::bench_itertools_processing ... bench:   7,020,214 ns/iter (+/- 311,371)
test tests::bench_std                  ... bench:     409,480 ns/iter (+/- 41,036)
test tests::bench_std_processing       ... bench:   9,092,307 ns/iter (+/- 484,857)

test result: ok. 0 passed; 0 failed; 0 ignored; 8 measured; 0 filtered out; finished in 42.75s
```

You're using `join` for cases with not more than 100 items and with no
processing at all (so `std` is even faster than the custom `Join`). IMO, it's
not worth it in this case. If you want to optimize go ahead and make a
flamegraph with this https://github.com/jonhoo/inferno and then we can focus on
the parts of the library that actually affect performance.

I also tried it back when it was `TryJoin` and in that case it was indeed much
faster.

kstep:
Hmm, OK. According to the benchmarks you provided, my version is almost twice as fast as std collect()+join() version (5.7*10^6 ns vs 9*10^6 ns), but still comparable to fold() version. If you want to drop the Join impl, what about replacing it with fold() version instead of collect()+join()?

Also, I must point out, while it's usually a small number of ids to process, there can be many queries (e.g. I update playback status every several seconds, there can be other examples as well), and also it's not just about time, it's also about memory footprint.

marioortizmanero:
> Hmm, OK. According to the benchmarks you provided, my version is almost twice as fast as std collect()+join() version (5.7*10^6 ns vs 9.0*10^6 ns), but still comparable to fold() version. If you want to drop the Join impl, what about replacing it with fold() version instead of collect()+join()?

Yes, but the `Join` implementation is only faster when there's processing added to it (like the `map`). I only made the processing benchmarks because we needed something similar with the `TryJoin` implementation (in which case `custom` was faster in both cases). In the latest `Id` version, the `std` (409,480 ns/iter) solution is actually faster than `custom` (599,477 ns/iter) in our current use case, which is just `ids.iter().collect().join()`.

> Also, I must point out, while it's usually a small number of ids to process, there can be many queries (e.g. I update playback status every several seconds, there can be other examples as well), and also it's not just about time, it's also about memory footprint.

I understand your point but when there are many queries it also affects other parts of the library. What I mean is that this part of the program is only a .001% (completely made up) of the execution time, and that we should focus on optimizing the rest. I would say that the number of queries doesn't matter much when considering what to optimize. About the memory footprint benchmarks, I'm not sure how I could benchmark that easily with Rust. Any Ideas?

kstep:
OK, I remove the `Join` trait altogether. To reduce boilerplate I introduced `join_ids()` function instead, it's implementation really doesn't matter, it here just to keep code DRY.

marioortizmanero:
Great job! Some questions/ideas:

* I don't really understand the `Sealed` trait in `idtypes`. What's that for?
* Why do we need the `IdBuf` type again? I still think using `String` isn't that much of a performance impact in `Id` (and we could easily use `Id<Artist>` instead of having to do `type ArtistId<'a> = Id<'a, Artist>`), but even then I don't think we need `IdBuf`.
* What about changing `from_id_or_uri` to just `new`? Just an idea.
* I would change `_type` in `Id` to something like `kind` to avoid the weird underscore

kstep:
`IdBuf` is just a convenience type for end-users. I tried to use the new API in my own project, and sometimes it's convenient to work with owning types (especially while working in a multi-threaded environment, when you need to pass data between threads via channels). We can drop it here, but end-users will have to create their own types. The same goes for `TypeId<'a>` aliases for `Id<'a, Type>`s: it's more convenient to import these types in a bunch than `Album`/`Track`/... types, as they may conflict with other names (e.g. names from `Type` variants). Pure practical types. We sure can live without them, it's just a little bit easier to live with them.

The `Sealed` trait is here to stop end-users from implementing `IdType` trait for their own types. It's a common pattern (see also: https://docs.rs/sealed/0.1.3/sealed/, https://rust-lang.github.io/api-guidelines/future-proofing.html), it came from Scala world. Without it, our end-users can crate their own id types, which would be unexpected for our method and can lead to unpredictable results.

marioortizmanero:
> IdBuf is just a convenience type for end-users. I tried to use the new API in my own project, and sometimes it's convenient to work with owning types (especially while working in a multi-threaded environment, when you need to pass data between threads via channels). We can drop it here, but end-users will have to create their own types. Also, it's a common practice (e.g. in std) to have owning/borrowing types in pairs (Path/PathBuf, CStr/CString, str/String, ...). I thought it would be nice to provide these types for end-users, as they as well may want to have them anyway, or we risk having a lot of similar owning types re-implemented in a lot of dependent crates.

Seems good to me then! I'd keep it as well.

> The same goes for TypeId<'a> aliases for Id<'a, Type>s: it's more convenient to import these types in a bunch than Album/Track/... types, as they may conflict with other names (e.g. names from Type variants, or the model::show::Show type). Pure practical types. We sure can live without them, it's just a little bit easier to live with them. By having these aliases we can still do use rspotify::model::* and have Id types along with other types without fear of conflicts.

Yeah that makes sense, I like the idea.

> The Sealed trait is here to stop end-users from implementing IdType trait for their own types. It's a common pattern (see also: https://docs.rs/sealed/0.1.3/sealed/, https://rust-lang.github.io/api-guidelines/future-proofing.html), it came from Scala world. Without it, our end-users can create their own id types, which would be unexpected for our method and can lead to unpredictable results.

Okay I completely forgot about that. Could you add a quick reference/explanation in the code for future readers who might not know about it?

> The _type field is named this way in fashion with other places: there are a lot of _type fields all over the code. It's named this way to keep code style unified.

Yeah you're right. I can see that other structs in the models use the same name. I'm not a big fan of it but whatever.

---

Since this changes so much code I'll give it a second review sometime soon so that we can finally merge this. @ramsayleung, any thoughts?

src/client.rs
```diff
@@ -2131,6 +2033,18 @@ impl Spotify {
     }
 }
 
+fn join_ids<'a, T: IdType>(ids: impl IntoIterator<Item = Id<'a, T>>) -> String {
```
marioortizmanero, line 2036: Why not just use `collect` and `join(",")`? We have to allocate anyway with this implementation.

kstep:
I'm back from the Christmas holidays and here's another bunch of changes.

- Make `Offset` type more light-weight and type-safe. Also removed `Deserialize` and `Serialize` traits impl from it, as it's not used anywhere anyway. Also, I turned it into an enum, as it's impossible to set both position and uri as an offset anyway.
- Simplified `join_ids` helper function.

ramsayleung:
I was busy on job and other pull requests before, Sorry for not reviewing earlier :)

It's a big pull request, I am trying to understand your design.

> IMO passing Id to the endpoints instead of &str is beneficial because the user can choose how to parse it when they know whether it's an ID or a URI. It's also a bit more type-safe.

yes, I agree with Mario's point, it would be better to pass `Id` type instead of `&str` by leverage Rust's type system, making it a bit more type-safe.

> my assumption would be that Rspotify aims to be the most flexible and feature-full one, which isn't fully compatible with being the most performant one.

Totally agree! Rspotify aims to be the most flexible ,feature-full and the ergonomic one. Then, if we want to improve the performance, we must have a benchmark for Rsptofy firstly, in order to find out the performance bottleneck. I know there are a lot of unnecessary allocation in Rpsotify, but we are not sure if it's the bottleneck.

src/model/idtypes.rs
```diff
@@ -0,0 +1,300 @@
...
+    #[serde(default)]
+    _type: PhantomData<T>,
+    #[serde(flatten)]
+    id: String,
+}
+
+impl<T> AsRef<Id<T>> for IdBuf<T> {
+    fn as_ref(&self) -> &Id<T> {
+        // Safe, b/c of the same T between types, IdBuf can't be constructed from invalid id, and Id is just a wrapped str with ZST type tag
+        unsafe { &*(&*self.id as *const str as *const Id<T>) }
```
ramsayleung, line 117: If it's safe, is it fine to remove `unsafe` block ?

src/model/idtypes.rs
```diff
@@ -0,0 +1,300 @@
...
+impl private::Sealed for Playlist {}
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum User {}
+impl private::Sealed for User {}
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Show {}
+impl private::Sealed for Show {}
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum Episode {}
+impl private::Sealed for Episode {}
```
ramsayleung, line 85: These `impl IdType for xxType`, `xxxId`, `xxxIdBuf`, `enum xxx` statements are a bit verbose, is there a way to define them cleanly? for example, use a `Macro` to do so?

src/model/idtypes.rs
```diff
@@ -0,0 +1,300 @@
...
+
+/// A Spotify object id of given [type](crate::model::enums::types::Type)
+///
+/// This is a not-owning type, it stores a &str only.
+/// See [IdBuf](crate::model::idtypes::IdBuf) for owned version of the type.
+#[derive(Debug, PartialEq, Eq, Serialize)]
+pub struct Id<T> {
+    #[serde(default)]
+    _type: PhantomData<T>,
```
ramsayleung, line 94: Sorry for not getting your point, I am a little bit confused about the usage of `PhantomData<T>`, what kind of problem does it solve, is it fine to remove `PhantomData<T>` or replace it with another type?

src/model/idtypes.rs
kstep: The unsafe block doesn't mean the code is unsafe. It's needed here to let me do raw pointers type-casting and dereferencing. I can't do it without the `unsafe` block. It really means: “Hey, compiler, I know, you can't prove the code is safe, but I, a human programmer, do know better in this concrete time, I proved it to be safe, so keep calm, I take responsibility for all the type and memory checks”.

src/model/idtypes.rs
kstep: All generic type arguments must be bound to some struct field. It's a hard Rust requirement. Here type argument is just a marker, there's no useful field of this type in the structure. That's why we need zero-sized dummy types like `PhantomData<T>`. Its role is to show the compiler I really use this type argument `T`.

src/model/idtypes.rs
kstep: Sure, I can use either a macro or a [sealed](https://crates.io/crates/sealed) crate here.

src/model/idtypes.rs
kstep: You can read about `PhantomData` here: https://doc.rust-lang.org/nomicon/phantom-data.html

src/model/idtypes.rs
marioortizmanero: Yes and it seems safe because `Id`'s only field is a `str`, and you seem to be doing `String -> str -> Id`. Doesn't this require `#[repr(C)]` though? Or I guess that's only for cases where there's padding between members? Also, why not just use a [transmute](https://doc.rust-lang.org/nomicon/transmutes.html)?

marioortizmanero (commented):
I'd love to get this merged next! I really like how you just use `Id` as `&Id` instead of `Id<'a>`. Love the docs as well, good job.

What's left:

* Can you keep the comments to 80 characters long?
* Can you fix the merge conflicts? If you need any help with these, let me know.

src/model/idtypes.rs
kstep: I started with `transmute()` version, but then clippy told me I better use type-casting here. Also, transmute is an unsafe function as well, so using transmute doesn't give any benefits. `#[repr(C)]` here is unnecessary, because this struct has only one real field, which is a noop in Rust, so after compilation it has the same real memory layout as `str`.

kstep:
I resolved the conflicts, cleaned and reformatted the code a little bit. Please feel free to check if it's OK now.

src/model/idtypes.rs
```diff
@@ -0,0 +1,289 @@
...
+pub struct IdBuf<T> {
+    #[serde(default)]
+    _type: PhantomData<T>,
+    #[serde(flatten)]
+    id: String,
+}
+
+impl<T> AsRef<Id<T>> for IdBuf<T> {
+    fn as_ref(&self) -> &Id<T> {
+        // Safe, b/c of the same T between types, IdBuf can't be constructed from invalid id, and Id is just a wrapped str with ZST type tag
```
marioortizmanero, line 89: You missed this comment, which is too long

src/model/idtypes.rs
```diff
@@ -0,0 +1,289 @@
...
+        unsafe { &*(&*self.id as *const str as *const Id<T>) }
+    }
+}
+
+impl<T> Borrow<Id<T>> for IdBuf<T> {
+    fn borrow(&self) -> &Id<T> {
+        // Safe, b/c of the same T between types, IdBuf can't be constructed from invalid id, and Id is just a wrapped str with ZST type tag
```
marioortizmanero, line 96: You missed this comment, which is too long

marioortizmanero (changes requested):
Beside the small nitpicks I commented, LGTM! What do you think, @ramsayleung?

CHANGELOG.md
```diff
@@ -130,6 +131,13 @@ If we missed any change or there's something you'd like to discuss about this ve
 	- `user_follow_users`
 	- `user_unfollow_users`
 	- `audios_features`
+  + The endpoints which changes parameter from `String` to `&Id<Type>`:
+        - `get_a_show`
+        - `get_an_episode`
+        - `get_shows_episodes`
+  + The endpoint which changes parameter from `Vec<Map<String, Value>>` to `Vec<TrackPositions>`:
+        - `playlist_remove_specific_occurrences_of_tracks`
+- The `Offset` type is now an enum to match API logic, `Offset::Position` is `u32` now (it's not a position in time, it's a position in a playlist, and you can't have both `position` and `uri` fields at the same time).
```
kstep, line 140: If you try to post something like this (using test console: https://developer.spotify.com/console/put-play/):

```json
{
  "context_uri": "spotify:album:5ht7ItJgpBH7W6vJ5BqpPr",
  "offset": {
    "position": 5,
    "uri": "spotify:track:4HbKgKkVeRvpL6yKlRf94E"
  },
  "position_ms": 0
}
```

You get an error:
```json
{
  "error": {
    "status": 400,
    "message": "Illegal offset: should be position XOR uri"
  }
}
```

So this `Offset` object is really an enum. Also, the `position` field is not a time in a song (there's a separate `position_ms` field in the request body for it), it's a position in a play context (e.g. a playlist or an album), so it's really a number, not some duration.

marioortizmanero: Good catch!

src/model/idtypes.rs
```diff
@@ -0,0 +1,299 @@
...
+            }
+        )+
+    }
+}
+
+sealed_types!(Artist, Album, Track, Playlist, User, Show, Episode);
+
+impl PlayContextIdType for Artist {}
+impl PlayContextIdType for Album {}
+impl PlayableIdType for Track {}
```
ramsayleung, line 40: I'm a little bit confused about these two types, what's it for?

ramsayleung:
> Beside the small nitpicks I commented, LGTM! What do you think @ramsayleung 

Yep, it's same to me, beside the small nitpicks, it looks good to me now :)

src/model/idtypes.rs
marioortizmanero: These seem to be for generics, so that you can accept multiple types of related IDs (playable types such as tracks/episodes, etc).

kstep: Yes, you are correct. Some methods accept only track/episode ids (`PlayableIdType`), and some methods operate on play contexts (`PlayContextIdType`). Hence these traits used to limit a set of id types to accept.

src/model/idtypes.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// - `IdError::InvalidFormat` - if it can't be splitted into type and
+    ///    id parts.
+    pub fn from_uri<'a, 'b: 'a>(uri: &'b str) -> Result<&'a Id<T>, IdError> {
+        let rest = uri.strip_prefix("spotify").ok_or(IdError::InvalidPrefix)?;
+        let sep = match rest.chars().next() {
+            Some(ch) if ch == '/' || ch == ':' => ch,
+            _ => return Err(IdError::InvalidPrefix),
+        };
+        // It's safe to do .get_unchecked() because we checked the first char above
+        let rest = unsafe { rest.get_unchecked(1..) };
```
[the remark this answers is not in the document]
kstep: It is, but there will be bounds check then. Bounds check is not really needed here, as we already checked the first char is 1-byte above.

src/model/idtypes.rs
```diff
@@ -0,0 +1,299 @@
...
+    #[serde(default)]
+    _type: PhantomData<T>,
+    #[serde(flatten)]
+    id: str,
+}
+
+/// A Spotify object id of given [type](crate::model::enums::types::Type)
+///
+/// This is an owning type, it stores a String.
+/// See [IdBuf](crate::model::idtypes::Id) for light-weight non-owning type.
```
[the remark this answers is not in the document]
kstep: Good catch! Thank you!

src/model/idtypes.rs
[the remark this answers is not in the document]
ramsayleung: I get your point, IMHO I think these bounds check doesn't cost much resources. 

As for the `unsafe` statement, my preference is that if we don't need it, we don't have to keep it. And the `reset.get_unchecked(1..)` statement is safe **for now**, it depends on the code context that we checked the first char above. 

As the codebase changed, the code context will be changed or not, then safe prerequisite will be broken.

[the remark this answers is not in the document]
kstep: Hm, ok. I changed the code so it both doesn't require bounds check, and doesn't have unsafe blocks.

kstep:
Tests started to fail on the authentication stage. I'm pretty sure, it's unrelated to my changes. It's either some transient error, or some auth token got invalid for some reason, or an API got changed in some way.

src/model/idtypes.rs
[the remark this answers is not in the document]
ramsayleung: Great job! :)

marioortizmanero (approved):

marioortizmanero:
I'm merging the PR then! Thanks a bunch for the changes and for your patience, @kstep, it's been a long time haha. If you want to change anything just make a new PR as this one is huge already.

As to the Auth issue, I'll take a look later then.

ramsayleung:
Cheers!

Issue #135, opened 2020-10-10
Consider making the cache file a feature 


marioortizmanero:
The cache file feature is something I'm not sure how I feel about. While it's definitely useful, I'm not sure if it should be the default behaviour. It could be moved into a feature, which when activated, would use the cache file anywhere possible. Or instead of having `do_something` and `do_something_without_cache`, it could be `do_something` and `do_something_with_cache`.

What do you think? I need some opinions on this. I don't actually plan on changing anything but I want to propose any breaking changes before the next release.

ramsayleung:
Yes, I think it's more proper to have a switch to control how does it work. My original thought was using `cache_path` as a trigger, as what I commented in #129:

```rust
    #[builder(default = "PathBuf::from(\".spotify_token_cache.json\")")]
    pub cache_path: PathBuf,
	// I am thinking about whether it's better to make cache_path optional, so we can decide to whether save the token into file or not by checking if cache_path exists, so we can remove function get_token_without_cache
``` 
Now, probably it's more suitable to make it as a feature, switching on compile time instead of runtime, as what you suggests.

ramsayleung:
Any other proposals? I am gonna to start to work on this feature on this weekend as what @marioortizmanero suggests, makes cache file as a new feature.

marioortizmanero:
Can you cross off more things from the meta issue before this? I would wait a bit more before this, let's see if anyone opposes this.

ramsayleung:
Sure! We could leave this issue alone for a while.

marioortizmanero:
This has been inactive for a while and noone has left their opinion so I would say we can introduce the `cache-file` feature. Should it be activated by default?

Also, do you still want to work on it @ramsayleung?

ramsayleung:
> Should it be activated by default?

I think it would be good to active this feature by default, just makes `rspotify` work as before.

> Also, do you still want to work on it @ramsayleung?

Not yet, I am working on this issue: https://github.com/ramsayleung/rspotify/issues/163, I would like to work on this feature after merging the PR of #163, and we could say we have crossed off most of TODO items in #127

Issue #124, opened 2020-09-09
Add unlimited endpoints


marioortizmanero:
One thing I really like about [tekore](https://github.com/felix-hilden/tekore) is that it offers endpoints without limits. Spotify has a limit of items that may be returned with certain requests, such as the followed artists or the songs in a playlist.

There are two different ways to do that with tekore AFAIK:
* [scoped for a number of calls](https://tekore.readthedocs.io/en/stable/reference/client.html#tekore.Spotify.max_limits)
* [when initializing the client](https://tekore.readthedocs.io/en/stable/reference/client.html#tekore.Spotify)

And then the endpoints themselves are what you'd expect: [`followed_artists`](https://tekore.readthedocs.io/en/stable/reference/client.html#tekore.Spotify.followed_artists). These don't mention the "unlimited" mode, it's only tied to the client itself. The client will internally send multiple requests until the endpoint returned value is complete, and return the results.

How this could be implemented in this library could of course vary, but this feature itself is very useful and interesting.

PR #166, opened 2020-12-16
[WIP/Proposal] Add Stream/Iterator based versions of methods with paginated results

icewind1991:
Note, this currently only implements the streaming version for one method to avoid wasting work until we're certain of the design.

## Description

Adds an alternative version of methods that return paginated results, providing instead an `Iterator` or `Stream` (for `sync`/`async` respectively)

## Motivation and Context

Manually iterating over paginated results is pretty annoying to do, by providing an streaming alternative we can make this a lot easier for consumers of the api.

## Dependencies 

New dependencies:

- for async:
   - [async-stream](https://crates.io/crates/async-stream) for creating the stream from async method calls
   - [futures](https://crates.io/crates/futures) provides the `Stream` trait

## Type of change

- New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

Added examples using the new methods and running them


marioortizmanero:
Hi! Thanks a lot for the contribution.

I definitely like the idea. It's a big step towards https://github.com/ramsayleung/rspotify/issues/124. Here are a few questions:

* We've been trying to avoid the `itertools` dependency for a while to try to reduce the number of dependencies. Would it be acceptable to not use it? At this point it might be worth importing, but I'd rather not.
* AFAIK `async-stream` (or `futures-util`, not sure), just like `async-trait`, is something that may be replaced in the future by the standard library, right? All I've found about this is a blog post from withoutboats: https://boats.gitlab.io/blog/post/for-await-i/ and the GAT tracking issue: https://github.com/rust-lang/rust/issues/44265. I would prefer a more native approach if possible.
* Shouldn't the `50` used to continue the pagination be more flexible? Some endpoints return more results than other so we should be able to tweak it.
* Should we straight up remove the manual methods for paginating? IMO there's no point in using them after this.

icewind1991:
- removed `itertools`
- replaced `futures-util` with `futures` since that is the actual source of the `Stream` trait, `futures-util` remains a dev-dependency for using the stream in the example
- I don't think GAT makes `async-stream` irrelevant, afaik, the alternative would be to manually implement `Stream::poll_next`, either way since the api only specifies `impl Stream` the specific stream implementation can be replaced later if a better approach is found without breaking compatibility.
- removing the paged api makes sense to me, only reason to keep them would be for backwards compatibility

kstep:
I'd prefer to have paginated API reachable. Sometimes it's good to have more low-level control over API calls than have an abstraction.

icewind1991:
- Switched to using a trait "alias" to prevent having to write duplicate async/sync versions of the streaming methods

ramsayleung:
It's a great idea to have some new ergonomics API, but I have it's reasonable to keep the paginated API, since some developers use `rspotify` to develop Web Application, they probably need to store the track and album information into database, and the paginated information is like an anchor for those Web Application to tell them where it's start point.

Maybe, we could add a new feature named `streamify`(or other names) to enable stream/iterator based version of methods with paginated result. We could keep them both, and offer different choices for different developers.

marioortizmanero:
I agree. Maybe this could be feature-gated. That way we can avoid the `_stream` prefix, kinda like what I want to do with the cache files at #135. How about a `streams` feature?

marioortizmanero:
I've updated this PR to include the latest changes in `master`. We should work on merging it before continuing with others to avoid having stale PRs.

Also, I think the module/feature name should be `pagination`, if anything. Otherwise `streams` is confusing because that's supposed to be only the async implementation. `pagination` is more broad and works for both streams and iterators.

Cargo.toml
```diff
@@ -54,7 +58,7 @@ reqwest-native-tls-vendored = ["reqwest/native-tls-vendored"]
 ureq-rustls-tls = ["ureq/tls"]
 
 # Internal features for checking async or sync compilation
-__async = ["async-trait"]
+__async = ["async-trait", "async-stream", "futures"]
```
ramsayleung, line 61: Is `futures-util` crate necessary for `__async` feature, or we could remove `futures-util` from dependencies list and leave it in `dev-dependencies`?

src/client.rs
```diff
@@ -933,6 +934,19 @@ impl Spotify {
         self.convert_result(&result)
     }
 
+    /// Get a list of the songs saved in the current Spotify user's "Your Music"
+    /// library.
+    ///
+    /// [Reference](https://developer.spotify.com/web-api/get-users-saved-tracks/)
+    pub fn current_user_saved_tracks_stream(
```
ramsayleung, line 941: After this PR merged, what's the next step? Planing to add a new method with `_stream` suffix for each endpoint except `current_user_saved_tracks`?

marioortizmanero: That's what I was explaining and questioning in this comment - https://github.com/ramsayleung/rspotify/pull/166#issuecomment-793698024

Seeing as there isn't anyone else giving their opinion I would just come to an agreement between you and I and just use that.

src/pagination/iter.rs
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    pages.flat_map(|result| ResultIter::new(result.map(|page| page.items.into_iter())))
+}
+
+/// Iterator that repeatedly calls a function that returns a page until an empty
+/// page is returned.
+struct PageIterator<T, E, Function>
+where
+    T: Unpin + 'static,
```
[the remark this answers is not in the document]
marioortizmanero: Ok, considering this iterator should only be implemented inside Rspotify, I can simplify it further to remove `Unpin` and use `rspotify::Error` instead o `std::Error`

src/client.rs
ramsayleung: I prefer `_stream` suffix to `_auto` suffix, since the latter one is not that self-explanatory as `_stream` suffix, and `stream` is defined as a sequence of elements supporting sequential and parallel aggregate operations in a we-all-know way, such as Java's `Stream` and Rust's Future `stream`.

And let me know what you think now :-)

marioortizmanero: Yes but the thing is that `_stream` might be confusing, as it returns an `Iterator` when sync mode is enabled, and it returns a `Stream` when async mode is enabled. `_auto` works for both, and so does `Paginator`, IMO. Perhaps there's a better alternative than `_auto`, but my concern was that we were mixing up the async implementation with the generalized one.

ramsayleung: Yes, I get you point! It makes sense. Perhaps there's indeed a better choice than `_auto`. But in this moment, no clue comes to my mind

ramsayleung: By the way, I am thinking about is it possible to generate the new method with `_stream` or `_auto` suffix by macro, since the code pattern is similar.

src/pagination/stream.rs
```diff
@@ -0,0 +1,34 @@
...
+    use async_stream::stream;
+    let mut offset = 0;
+    stream! {
+        loop {
+            let page = f(page_size, offset).await?;
+            offset += page.items.len() as u32;
+            for item in page.items {
+                yield Ok(item);
+            }
+            if page.next.is_none() {
```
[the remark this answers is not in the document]
marioortizmanero: Done! Simplified the implementation and added some tests

Cargo.toml
marioortizmanero: We only need it in `dev-dependencies`. Removed it from `dependencies`.

src/client.rs
marioortizmanero: > By the way, I am thinking about is it possible to generate the new method with _stream or _auto suffix by macro, since the code pattern is similar.

This is non-trivial to implement because of generic types in functions. A macro that takes these into account is quite complicated and not worth the effort in this case, unfortunately.

ramsayleung: There is another suffix name came to my mind, `_range` suffix, inspiration from C++ [Range library](https://en.cppreference.com/w/cpp/ranges)  :)

marioortizmanero: Here's another idea I added to my comment, which I modified: what if the endpoints are automatically paginated by default, and you can use the `_manual`-suffixed endpoints in case you want more control over them? It's a saner default, as most times you just want an iterator, and we solve the naming issue pretty easily :)

ramsayleung (approved):

src/client.rs
ramsayleung: Yes, it makes much sense. We could make the endpoints paginated by default, since the developers want iterator at most times as you point out. My concern is that if we change all manual endpoints to be paginated and then add a `_manual` suffix for the original one, then we will change the function signature of all endpoints, it does break the compatibility.

marioortizmanero:
Nooo @ramsayleung don't merge this! It was only a showcase of how it should work. I still had to implement the rest of the endpoints and use `_manual`.

marioortizmanero:
Anyway, I'll just open a new PR.

ramsayleung:
Ooooh, I though this PR is a prototype, and it's ready to merge, my fault :(

marioortizmanero:
Yup, don't worry. I'll finish everything there and after we're done with that we can finally work on the auth rewrite!

PR #167, opened 2020-12-17
explicitly forbid having both client features enabled

icewind1991:
## Description

Instead of showing various `duplicate definition` errors when a user tries to compile the library with both client features, provide a meaningful error message explaining that both features can't be enabled at the same time.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,8 +154,9 @@
 //! ](https://github.com/ramsayleung/rspotify/tree/master/examples)
 //! which can serve as a learning tool.
 
-// disable all modules when both client features are enabled, this way only the compile error below gets
-// show instead of showing a whole list of confusing errors
+// disable all modules when both client features are enabled,
+// this way only the compile error below gets show
+// instead of showing a whole list of confusing errors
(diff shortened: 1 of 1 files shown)


ramsayleung (approved):

marioortizmanero (commented):
Great idea! And thanks for the contribution!

Can you try to split the lines so that they aren't longer than 80 characters? We try to keep comments at 80 and code at 120.

ramsayleung:
> We try to keep comments at 80 and code at 120. 

Yep, perhaps we could put it into a code style guide?

icewind1991:
done

marioortizmanero:
@ramsayleung I think a script isn't a viable solution because some lines need to be longer than 80 characters when there's a long URL and such. Perhaps we could mention it in the `Contributing` section inside the README.md. I'll make a PR.

Something like this could be added to the tests as well to make sure they aren't ran without the `env-file` feature. Otherwise you get errors like this, which make no sense if you aren't a developer:

```
thread 'test_album' panicked at 'called `Result::unwrap()` on an `Err` value: "`id` must be initialized"', tests/test_with_credential.rs:15:56
```

I'll make another PR for that and merge this.

PR #170, opened 2020-12-24
Mention line lengths in README

marioortizmanero:
## Description

This should fix what we discussed at #167 regarding line lengths that aren't checked automatically by `cargo fmt`. Let me know if you think it could be worded differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,8 @@ Please see the [changelog](./CHANGELOG.md) for a release history and indications
 
 If you find any problems or have suggestions about this crate, please submit an issue. Moreover, any pull request, code review and feedback are welcome.
 
+We use GitHub Actions to make sure the codebase is consistent (`cargo fmt`) and continuously tested (`cargo test`). We try to keep comments at a maximum of 80 characters of length (which isn't automatically checked by `cargo fmt`) and code at 120.
+
 ## Building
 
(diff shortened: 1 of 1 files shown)


ramsayleung:
It's a great point to point out the line lengths in `README.md`, but I am afraid of that other developers will miss the code guide part if they don't read the `README.md` carefully.

IMHO, it would be better to have a way to check line lengths automatically.

README.md
```diff
@@ -17,6 +17,8 @@ Please see the [changelog](./CHANGELOG.md) for a release history and indications
 
 If you find any problems or have suggestions about this crate, please submit an issue. Moreover, any pull request, code review and feedback are welcome.
 
+We use GitHub Actions to make sure the codebase is consistent (`cargo fmt`) and continuously tested (`cargo test`). We try to keep comments at a maximum of 80 characters of length (which isn't automatically checked by `cargo fmt`) and code at 120.
+
```
ramsayleung, line 21: Perhaps, we could have an `Code Guide` section?

marioortizmanero:
> IMHO, it would be better to have a way to check line lengths automatically.

As I pointed out in a different PR, this wouldn't be possible:

> I think a script isn't a viable solution because some lines need to be longer than 80 characters when there's a long URL and such.

README.md
marioortizmanero: done

ramsayleung (approved):

PR #171, opened 2020-12-25
Improved test error messages and small CI fix

marioortizmanero:
## Description

If you ran `cargo test` without the `env-file` feature by mistake or no credentials set, you'd get a somewhat weird error. This should provide more context for both `test_with_credential` and `test_with_oauth`. I also removed the `cli` feature from the CI tests as it's not really needed for now (CI can't interact with the user anyway).

## Type of change

## How Has This Been Tested?

I've just ran the tests a couple times expecting these errors

### DIFF
--- a/tests/test_with_oauth.rs
+++ b/tests/test_with_oauth.rs
@@ -41,7 +41,13 @@ pub async fn oauth_client() -> Spotify {
     } else if let Ok(refresh_token) = env::var("RSPOTIFY_REFRESH_TOKEN") {
         // The credentials must be available in the environment. Enable
         // `env-file` in order to read them from an `.env` file.
-        let creds = CredentialsBuilder::from_env().build().unwrap();
+        let creds = CredentialsBuilder::from_env().build().unwrap_or_else(|_| {
+            panic!(
+                "No credentials configured. Make sure that the required \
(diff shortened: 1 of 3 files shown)


tests/test_with_oauth.rs
```diff
@@ -41,7 +41,13 @@ pub async fn oauth_client() -> Spotify {
...
         // The credentials must be available in the environment. Enable
         // `env-file` in order to read them from an `.env` file.
-        let creds = CredentialsBuilder::from_env().build().unwrap();
+        let creds = CredentialsBuilder::from_env().build().unwrap_or_else(|_| {
+            panic!(
+                "No credentials configured. Make sure that the required \
+                environment variables are exported (`RSPOTIFY_CLIENT_ID`, \
+                `RSPOTIFY_CLIENT_SECRET`)"
+            )
+        });
```
ramsayleung, line 50: Perhaps we could notice the users that they could also configure credentials by enabling `env-file` feature, just make the error message as same as `test_with_credential.rs` ?

marioortizmanero: Done

ramsayleung (approved):

PR #176, opened 2021-01-10
Refresh token from cache when expired

borisfaure:
## Description

`read_token_cache` now refreshes token if token is expired.

## Motivation and Context

I wanted `spotify.prompt_for_user_token().await.unwrap();` to use the cache and refresh tokens

## Type of change

## How Has This Been Tested?

Tested manually with a token in cache that was expired for about 24h with this dumb code: https://github.com/borisfaure/spotifyctl

This *was* a draft PR since I wondered about the following:

- what to do about code duplication on refreshing the token
- whether to rewrite the cache here
- what to do on error writing the cache

Ps: This PR is no longer in draft state since it has been more than a month with no interaction from the maintainers: trying to get feedback without the PR being in Draft state.

### DIFF
--- a/src/oauth2.rs
+++ b/src/oauth2.rs
@@ -233,10 +233,27 @@ impl Spotify {
     pub async fn read_token_cache(&mut self) -> Option<Token> {
         let tok = TokenBuilder::from_cache(&self.cache_path).build().ok()?;
 
-        if !self.get_oauth().ok()?.scope.is_subset(&tok.scope) || tok.is_expired() {
+        if !self.get_oauth().ok()?.scope.is_subset(&tok.scope) {
             // Invalid token, since it doesn't have at least the currently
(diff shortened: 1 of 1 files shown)


borisfaure:
@ramsayleung, @marioortizmanero Can I please get your review on this?

marioortizmanero:
Hi! Yes, thank you very much for you PR. The thing is that we're still considering how the cached/refreshing tokens should work and be configured at #4.

What you included in this PR should only happen if the user has explicitly indicated that they want a self-refreshing token, and could be designed differently. This is specially important as more authentication methods are added, because what you implemented won't work for PKCE tokens, and it can quickly turn into spaghetti code.

You can help us out in there to take a decision on how this can be handled and then we'll implement it.

marioortizmanero:
Closing in favor of #223

Issue #96, opened 2020-05-27
Disable token caching


raffomania:
I'm writing a web app for multiple users and would like to save the API token in my database. Is there a way to prevent `.spotify_token_cache.json` from being created?

ramsayleung:
> Is there a way to prevent `.spotify_token_cache.json` from being created?

`.spotify_token_cache.json` is the default way to cache the token, there is no way to prevent to create `.spotify_token_cache.json` now :(

raffomania:
that option would be awesome!

ramsayleung:
I have completed this feature by adding two separate functions to get access token and refresh token without saving token into `cache_path` instead of adding a new option to disable cache. I think the current choice leads to less change of the library. If you want to know how to use the two new functions, check this example for more details :)
https://github.com/ramsayleung/rspotify/blob/92984d2bccaa9d9170f5b7962f1dcc2411b43212/examples/get_access_token_without_cache.rs#L22

https://github.com/ramsayleung/rspotify/blob/92984d2bccaa9d9170f5b7962f1dcc2411b43212/examples/get_access_token_without_cache.rs#L43

raffomania:
alright, seems like a larger API surface, but it solves my problem so I'm happy :)
Thanks a lot for making this crate and for the quick reply! <3

ramsayleung:
Thanks for your response, I am going to close this issue :)

Issue #173, opened 2020-12-28
Restructure the authentication process


marioortizmanero:
Back when I first rewrote the authentication process I did improve a few things I thought were necessary, like only having a single http client per `Spotify` instance, and being more ergonomic and easy to use.

But it still leaves a few things to be desired. For example, an `InvalidAuth` error variant is needed for whenever a user calls a method from `Spotify` that isn't available due to the authentication process. If one were to use the Client Credentials Flow, the endpoints with access to the user's data shouldn't be available (say `me` or `currently_playing_track`). You need to use the Authentication Code Flow for that. This worked well enough because we only had two different methods: the Client Credentials and the Authentication Code.

Now that the PKCE Code Authentication Flow has to be implemented, I think it's time to re-think the architecture we're following and make it more idiomatic. This auth flow can access the same endpoints the regular Code Authentication Flow can, but it requires a different setup.

We mainly need three levels for this problem:

* The HTTP client to actually make the requests, which already exists as `rspotify::http::BaseClient`. I'll call it `HTTPClient` from now on.
* `BaseClient`, a base Spotify client, with access to endpoints such as `tracks` or `search` that don't need access to the user's data. This uses `HTTPClient` to make requests.
* `UserClient`, an user authenticated Spotify client, with access to endpoints such as `me`. This may inherit from `BaseClient` and implement the extra endpoints over it, or use `HTTPClient` directly and only have the user-authenticated ones.

Rust doesn't really have OOP, but we can work it out with traits. Here's the first solution, where the `UserClient` inherits from `BaseClient` so that the Code Auth flows can access both `UserClient` and `BaseClient` methods:

![image](https://user-images.githubusercontent.com/25647296/103226309-1f73bf00-492c-11eb-9b17-f8d9bacd157e.png)

And here's another solution with a more horizontal hierarchy. `UserClient` only includes the user authenticated methods because it doesn't inherit from `BaseClient`, so the Code Auth flows have to implement both of them. I feel like this is more natural in Rust, and it seems more modular because it doesn't assume that user authenticated flows can access the regular ones.

![image](https://user-images.githubusercontent.com/25647296/103226918-75953200-492d-11eb-8309-bcf68042aaeb.png)

In both of these models the implementations take place in the traits, which only needs access to some `HTTPClient` methods, so the structs would only need to implement at most a couple getters for their internal fields, kinda like how `Iterator` works.

```rust
/// Just a rough idea of how it would look like to use the proposed models.
fn main() {
[...]

PR #178, opened 2021-01-16
Make cache file a optional feature

### LINKED ISSUE #135: Consider making the cache file a feature

ramsayleung:
## Description

Trying to fix #135 #96, make cache file as a optional feature.

## Motivation and Context

The cache file feature is useful, but the API surface is some kind of verbose, like `do_something` and `do_something_without_cache`. The developers have to explicitly choose `cache` function or `without_cache` function to control the way of how does the cache file work? It would be more ergonomic to make cache file as a feature to offer the right to developers to control how does it work ?

## Type of change

## How Has This Been Tested?

Test `cache-file` feature by hand:

+ `request_client_token`:

```rust
# Check if cache file is generated, expect to be nothing
cargo run --example track --no-default-feature --features="env-file cli client-reqwest request-default-tls" 

# Check if cache file is generated, expect that `.spotify_token_cache.json` is saved
cargo run --example track --no-default-feature --features="env-file cache-file cli client-reqwest request-default-tls"
```

+ `request_user_token`:

```rust
# Check if cache file is generated, expect to be nothing
cargo run --example me --features="env-file cli client-ureq ureq-rustls-tls" --no-default-features

# Check if cache file is generated, expect that `.spotify_token_cache.json` is saved
cargo run --example me --features="env-file cli client-ureq ureq-rustls-tls cache-file" --no-default-features
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -181,6 +181,9 @@ If we missed any change or there's something you'd like to discuss about this ve
   + Remove `SimplifiedPlayingContext`, since it's useless.
 - ([#177](https://github.com/ramsayleung/rspotify/pull/157)) Change `mode` from `f32` to `enum Modality`:
   + Change `AudioAnalysisSection::mode`, `AudioAnalysisTrack::mode` and `AudioFeatures::mode` from `f32` to `enum Modality`.
+- ([#178](https://github.com/ramsayleung/rspotify/pull/178)) Make cache file as a feature:
+  + Mark `refresh_user_token_without_cache`, `request_client_token_without_cache`, `request_user_token_without_cache` from public to private.
+  + Remove `prompt_for_user_token_without_cache`.
 
 ## 0.10 (2020/07/01)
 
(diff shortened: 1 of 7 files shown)


marioortizmanero:
I was thinking that if we end up accepting the architecture at #173 we could also make a trait for cached requests. That would avoid all these conditional expressions, which look quite ugly to me and make the code harder to follow and read. But I'm still not sure if fully trait-based code is  that common in Rust, I have to investigate more. 

What do you think?

ramsayleung:
>  That would avoid all these conditional expressions, which look quite ugly to me and make the code harder to follow and read. 

Yes, it looks ugly to me either, but it's the most straight way coming to my mind.

> But I'm still not sure if fully trait-based code is that common in Rust, I have to investigate more.

Great to see that, IMHO, fully trait-based code looks like interface implementation in OOP, I am also not sure that it's common in Rust

> Unfortunately I won't be able to make much progress on this until February because I'm on finals right now.

It's understandable, I will wait for your new progress :)

marioortizmanero:
I think we could try to implement without features, as I wouldn't want to have too many of them. Although they make the library more modular and less bloated, configuring them can be complicated and limits it to a compile-time configuration. The code can also get quite ugly this way (take this PR as an example, which is filled with `#[cfg]` statements.

We could make the Spotify client generic over a token type which implements a basic interface to interact with the token. The trait would be basically as you implemented it in this PR when the cache-file feature is enabled, which would be implemented by two structs:

* `Token`, which would act as a dummy
* `CacheToken`, which would read and save the file whenever necessary

Not sure how easy it is to implement this but it might be a nice idea imo. I'll look into it later on and try to implement it.

Issue #134, opened 2020-10-10
Optional parameters


marioortizmanero:
I would like to discuss how to handle optional parameters for this library. I recently made an article inspired by this issue which sums up the different ways to approach this: https://vidify.org/blog/rust-parameters/

I think the most convenient ones in there are the following:

**A**: Using a bare `Option<T>`  
**B**: Using `Into<Option<T>>`  
**E**: Endpoint-oriented interface. We have related groups, like `track`, `album`... which could be used as the endpoint implementations.  
**F**: Hybrid derive pattern. More complex but is super nice for the user, and we could find a way to make it easier to implement.  
**G**: There are optional parameters repeated lots of times throughout the code, like `market`, `country`, `limit`... So we might not even need the group implementation, just in the client.

Currently, we use the second one, but it's not consistent. Some functions also use the first one. We should decide one of them and use it consistently throughout the codebase.

My personal favorite is just the first one. Using `Into<Option<T>>` seems unnecessary to me and doesn't offer that much over a bare `Option<T>`. That, or approach **E**, which seems perfect for the parameters I mentioned: `market`, `country`... and we could just ignore it when these params are specified for an endpoint that doesn't need it. As long as it's well documented I don't think it should be a problem. `market()` would just set a value inside the client (similar to the constructor) which would be used for all the calls afterwards.

I'd like to tag @Koxiaet, who opened an issue with a very opinionated list of improvements; I'm sure they can help with this as well :)

Koxiaet:
I don't like option E, because if the user wants to call an endpoint that doesn't require the `market` parameter at best they'll have to choose a random dummy value (which adds unnecessary extra code) and at worst they'll get confused and spend time getting a proper value.

I think that approach F is too complex. Even if we had macros that made it easy to implement, it basically doubles the size of the library, it's not worth it.

Approach G is too limited; it doesn't allow for users to have different markets for different queries. We don't want to force the user to have multiple clients, ever.

Between options A and B I don't have a preference, they both work. I don't think it matters which one is chosen.

PS: what happened to options C and D?

marioortizmanero:
The thing with E is that if the endpoint doesn't need the parameter it doesn't have to be specified. Its value will only be used for endpoints that require it. But yeah, I'm skeptical about it as well. C and D seemed too basic to me and not exactly what we want. Of course, they can also be discussed.

[...]

PR #179, opened 2021-01-25
Fix not allowing from token as market/country

### LINKED ISSUE #28: "from_token" is currently not a valid parameter
Hey,

thanks for your work on the library. I'm currently trying to search for tracks using rspotify. According to the documentation, `from_token` can be passed as a `market` parameter to search in the market tied to the token.

However, the `search_*` functions seem to require an Option<Country> parameter. There seems to be no Country enum representing `from_token`, unless I am missing something?

Qluxzz:
## Description

Fixes "from_token" not being a valid option for market/country.

Create Market enum which can be a country or from_token in order to satisfy both conditions.

Update usages of Country to instead use Market enum.

Update mentions and parameters of country/market to be consistent with spotify API docs.

Fixes #28 

## Motivation and Context

To allow market to be automatically selected from the users access-token instead of having to be supplied manually.

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Please also list any relevant details for your test configuration:
I ran all tests but the most interesting ones would be the following since they used the old Country enum

- [X] ```Market::Country```: test_artist_top_tracks
- [X] ```Market::Country```: test_categories
- [X] ```Market::Country```: test_category_playlists
- [X] ```Market::Country```: test_recommendations
- [X] ```Market::Country```: test_search_artist

I've also added a test that uses Market::FromToken functionality, but since the market supplied is not static and changes with the oauth token credentials, it maybe wouldn't be such a good test?
- [X] ```Market::FromToken```: test_search_artist_with_from_token

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -320,14 +320,14 @@ impl Spotify {
     ///
     /// [Reference](https://developer.spotify.com/web-api/get-artists-top-tracks/)
     #[maybe_async]
-    pub async fn artist_top_tracks<T: Into<Market>>(
+    pub async fn artist_top_tracks(
         &self,
         artist_id: &str,
-        market: T,
+        market: Market,
     ) -> ClientResult<Vec<FullTrack>> {
         let mut params = Query::with_capacity(1);
 
-        params.insert("market".to_owned(), market.into().into());
+        params.insert("market".to_owned(), market.into());
 
         let trid = self.get_id(Type::Artist, artist_id);
         let url = format!("artists/{}/top-tracks", trid);
--- a/src/client.rs
+++ b/src/client.rs
@@ -229,7 +229,7 @@ impl Spotify {
 
         let mut params = Query::new();
         if let Some(market) = market {
-            params.insert("market".to_owned(), market.into());
+            params.insert("market".to_owned(), market.to_string());
         }
 
         let url = format!("tracks/?ids={}", ids.join(","));
@@ -303,7 +303,7 @@ impl Spotify {
             params.insert("offset".to_owned(), offset.to_string());
         }
         if let Some(market) = market {
-            params.insert("market".to_owned(), market.into());
+            params.insert("market".to_owned(), market.to_string());
         }
         let trid = self.get_id(Type::Artist, artist_id);
         let url = format!("artists/{}/albums", trid);
@@ -327,7 +327,7 @@ impl Spotify {
     ) -> ClientResult<Vec<FullTrack>> {
         let mut params = Query::with_capacity(1);
 
-        params.insert("market".to_owned(), market.into());
+        params.insert("market".to_owned(), market.to_string());
 
         let trid = self.get_id(Type::Artist, artist_id);
         let url = format!("artists/{}/top-tracks", trid);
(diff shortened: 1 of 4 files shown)


src/model/enums/misc.rs
```diff
@@ -91,3 +93,20 @@ pub enum Modality {
...
+/// Limit the response to a particular market
+///
+/// FromToken is the same thing as setting the market parameter to the user's country.
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub enum Market {
+    Country(Country),
+    FromToken,
+}
+
+impl Into<String> for Market {
```
Qluxzz, line 105: Should this be moved somewhere else in order to keep this file to only declarations of enums?

src/client.rs
```diff
@@ -316,20 +316,18 @@ impl Spotify {
     ///
     /// Parameters:
     /// - artist_id - the artist ID, URI or URL
-    /// - country - limit the response to one particular country.
+    /// - market - limit the response to one particular country.
     ///
     /// [Reference](https://developer.spotify.com/web-api/get-artists-top-tracks/)
     #[maybe_async]
-    pub async fn artist_top_tracks<T: Into<Option<Country>>>(
+    pub async fn artist_top_tracks<T: Into<Market>>(
```
marioortizmanero, line 323: You can remove this `Into<Market>` and just use `Market`, as `Market` doesn't implement methods like `From<String>` AFAIK. This `Into` was used previously to avoid `Some` when passing optional parameters, but this decision is being discussed at #134 anyway (you can take a look and leave your opinion if you like :))

src/model/enums/misc.rs
marioortizmanero: Nah I would say it's ok

marioortizmanero: As as side note: can you add a new line between the previous enum and this one? Thanks

marioortizmanero (commented):
I think the test you added is OK, since we can't check the results anyway. As long as it makes sure it doesn't break it's enough for now.

src/model/enums/misc.rs
```diff
@@ -91,3 +93,21 @@ pub enum Modality {
...
```
marioortizmanero, line 106: Why not use `ToString` instead of `Into<String>` by the way? That way you can use `market.to_string`, which:

* Avoids type inference issues (I saw a case where you had to declare `market: String` to let Rust know it's `Into::<String>`)
* `to_string` is more explicit and easier to understand for the reader
* The struct is borrowed instead of moved, so you can keep using it

Qluxzz: Good question, guess I didn't really think about it. I'll fix it.

ramsayleung:
Is this PR is ready to merge, it seems there is no more update after the last commit?

Qluxzz:
Yes, I implemented the to_string trait instead of into trait as @marioortizmanero asked me about, and then requested a review.

marioortizmanero:
Yes sorry! I was busy but I'm finally done with finals. Merging this, thanks for the contribution :)

marioortizmanero (approved):

Qluxzz:
@marioortizmanero no worries, hope you aced your finals!

PR #182, opened 2021-02-09
Fix broken reference link in documentation.

### LINKED ISSUE #180: Update documentation links
**Describe the bug**

The reference links in the docs point to the previous version. For instance, this is the link used in the docs:

https://developer.spotify.com/web-api/get-artists-top-tracks/

It now redirects to:

https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-an-artists-top-tracks

Which is fine, really. But some other endpoint links like the search one are broken. So I would just change all links to the new ones.

ramsayleung:
## Description

Update documentation links, fixed #180 

## Motivation and Context

Some links used in the docs are broken, and some links are redirect to new url, so just update them to the latest one.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

Open updated links with browser, then check if these links are ok.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -61,12 +61,12 @@ pub type ClientResult<T> = Result<T, ClientError>;
 /// API as part of the JSON response object.
 #[derive(Debug, Error, Deserialize)]
 pub enum APIError {
-    /// See https://developer.spotify.com/documentation/web-api/reference/object-model/#error-object
+    /// See https://developer.spotify.com/documentation/web-api/reference/#object-errorobject
     #[error("{status}: {message}")]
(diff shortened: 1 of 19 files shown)


ramsayleung:
It seems that the previous pull request which has been merged into master leads to the failure of the CI in this pull request

https://github.com/ramsayleung/rspotify/runs/1849138761

marioortizmanero (approved):
Hey thanks for the PR! Yes, I think we can merge this for now, and I'll try to investigate what's going on with the CI. It looks like what happened when we upgraded to Tokio 1.0 but it wasn't supported yet.

PR #185, opened 2021-02-14
Polish Token and OAuth

ramsayleung:
## Description

As @Kestrer pointed out in #127, the `expires_at` and `expires_in` are not that easy to use:

> TokenInfo stores expires_in as a u32, and additionally has the expires_at field. This is not easy to use, instead it should have a single expires field with the type `Instant`

There is no way to store expires information into a signle field, since the response of requesting to `Spotify` for an access token looks like this:

```json
{
   "access_token": "NgCXRK...MzYjw",
   "token_type": "Bearer",
   "scope": "user-read-private user-read-email",
   "expires_in": 3600,
   "refresh_token": "NgAagA...Um_SHo"
}
```

It's possible to deserialize this `expires_in` to `Token.expire` field, like this:

```rust
    pub fn deserialize<'de, D>(deserializer: D) -> Result<Instant, D::Error>
    where
        D: Deserializer<'de>,
    {
        let elapsed = u32::deserialize(deserializer)?;
        let duration = Duration::from_secs(elapsed as u64);
        let now = Instant::now();
        let instant = now
            .checked_sub(duration)
            .ok_or_else(|| Error::custom("Erreur checked_add"))?;
        Ok(instant)
    }
``` 

In order to convert expires_in(`elapsed`) to `Instant`, it's necessory to instantiate a `Instant` object in `deserialize` function, it turns out to be a stateful function. If we serialize the `Token` into cache file and deserialize it to `Token`, then the `expires` value will change unexpectedly.

Hence, it's necessary to keep two fields(`expires_in`, `expires_at`) to handle expire information, the `expires_in` has been updated from `u32` to `std::time::Duration`, and the `expires_at` field has been updated from `u64` to `chrono::DateTime`.

For more details about the support of `serde` for `std::time::Instant`, check this issue: https://github.com/serde-rs/serde/issues/1375

## Motivation and Context

To make `rspotify` ergonomic to use. Modification:

+ Change `Token.expires_in` from `u32` to `std::time::Duration`
+ Change `Token.expires_at` from `i64` to `chrono::DateTime<Utc>`
+ Change `Token.scope` from `String` to `HashSet`.
+ Change `OAuth.scope` from `String` to `HashSet`.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

- [x] `test_write_token`
- [x] `test_token_is_expired`


src/oauth2.rs
```diff
@@ -210,8 +258,8 @@ impl Spotify {
             .post_form(auth_urls::TOKEN, Some(&head), payload)
             .await?;
         let mut tok = serde_json::from_str::<Token>(&response)?;
-        tok.expires_at = Some(datetime_to_timestamp(tok.expires_in));
-
+        tok.expires_at = Utc::now()
```
marioortizmanero, line 261: This is a bit confusing, what are you trying to do?

marioortizmanero (commented):
Great job! I've just included a few commits to have neater iterators for converting between HashSet and Vec, and a few other improvements, although if you consider any of them let me know and revert them.

src/oauth2.rs
```diff
@@ -7,44 +7,98 @@ use serde::{Deserialize, Serialize};
...
+    use serde::{de, Deserialize, Serializer};
+    use std::time::Duration;
+
+    /// Deserialize `std::time::Duration` from milliseconds (represented as u64)
+    pub(in crate) fn deserialize<'de, D>(d: D) -> Result<Duration, D::Error>
+    where
+        D: de::Deserializer<'de>,
+    {
+        let duration: u64 = Deserialize::deserialize(d)?;
+        Ok(Duration::from_secs(duration))
```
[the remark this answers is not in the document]
ramsayleung: Oooh, it's a mistake, they both should be `seconds`, since the `expires_in` field is represented in second

src/oauth2.rs
ramsayleung: `Rspotify` use two fields to represent the expiration information, the `expires_in` field is used to store the time period for which the access token is valid(for example, 3600s), then the `expires_at` field is used to store the expiry datetime. For example, the current time is `2021-02-15 00:00:00` in Utc timezone, then the `expires_in` is `3600s`, then expiry datetime `expires_at` is `2021-02-15 01:00:00`. The token is valid if it's before `2021-02-15 01:00:00`.

If you are confusing about this field, probably we need a better, self-explanatory name for this field instead of `expires_at`.

ramsayleung:
>  I've just included a few commits to have neater iterators for converting between HashSet and Vec, and a few other improvements, although if you consider any of them let me know and revert them.

You do a great job to reduce unnecessary operation, it's no way to revert them :)

src/oauth2.rs
```diff
@@ -58,24 +59,17 @@ mod space_separated_scope {
             scope
                 .split_whitespace()
                 .map(|x| x.to_owned())
-                .collect::<Vec<String>>(),
```
ramsayleung: IMHO, it's unnessary to collect iterator into a Vec explicitly, the `from_iter` can construct from an iterator, but `cargo clippy` failed for missing the `collect` statement:

```log
error: usage of `FromIterator::from_iter`
Error:   --> src/oauth2.rs:58:12
   |
58 |           Ok(HashSet::from_iter(
   |  ____________^
59 | |             scope.split_whitespace().map(|x| x.to_owned()),
60 | |         ))
   | |_________^ help: use `.collect()` instead of `::from_iter()`: `scope.split_whitespace().map(|x| x.to_owned()).collect()`
   |
   = note: `-D clippy::from-iter-instead-of-collect` implied by `-D warnings`
   = help: for further information visit rust-lang.github.io/rust-clippy/master/index.html#from_iter_instead_of_collect

```

https://github.com/ramsayleung/rspotify/pull/185/checks?check_run_id=1899884101

marioortizmanero: Ahhh I get it now. Yes, very smart. Give me a sec, I think I understand it.

marioortizmanero: Done! Take a look and let me know.

ramsayleung: Great job :)

marioortizmanero: Me too! :)

ramsayleung:
@marioortizmanero you could take time to review this PR again, I think the required commits are completed :)

src/oauth2.rs
marioortizmanero: Sorry, what I meant was that I didn't understand these two lines of code. They seem a bit obfuscated in my opinion, why is `.checked_add_signed(tok.expires_in);` not possible?

marioortizmanero (commented):
Ok after these comments it should be good to go!

Another thing I would like to consider is whether we'd be better off using `BTreeSet` instead of `HashSet` or not. Here's a comparison I've read: https://stackoverflow.com/questions/62558996/what-is-the-time-complexity-of-iterating-btreeset-and-hashset. This is also interesting: https://doc.rust-lang.org/std/collections/index.html#when-should-you-use-which-collection

What we mainly need is:

* Iteration of the elements
* Inserting new elements

It does look like a `HashSet` is the best way to go, but I just wanted to make sure.

src/oauth2.rs
ramsayleung: Yep, the original solution is impossible, since the required parameter type of `checked_add_signed ` is `chrono::Duration`, the type of `expires_in` is `std::time::Duration`.

So I just change the type of `expires_in` to `chrono::Duration`, it works.

ramsayleung: fixed: [`5893ed7` (#185)](https://github.com/ramsayleung/rspotify/pull/185/commits/5893ed770caef555dc466e865f24488999bdc5ee)

marioortizmanero (approved):
LGTM! Great work

PR #189, opened 2021-02-27
Add unit tests for pure methods

ramsayleung:
## Description

Add unit tests for pure methods like `get_uri` and `get_type` methods without external dependencies

## Motivation and Context

Add more unit tests for `rspotify` to make it solid and easy to maintain and refactor

## Dependencies 

None

## Type of change

Add unit tests

## How Has This Been Tested?

Check if all added tests pass


ramsayleung:
For unknown reason, the `test_read_token_cache` test will fail occasionally. I am trying  to figure out the reason.

marioortizmanero:
Maybe we should do something to make the scopes easier to use? Perhaps a `scopes!` macro would be easier to use than the whole `scope.split_whitespace().map(|x| x.to_owned()).collect()` boilerplate: `scope = scopes!("asd", "test")`.

In a different PR, of course, but I noticed while reading the changes in this one.

ramsayleung:
> Perhaps a scopes! macro would be easier to use than the whole `scope.split_whitespace().map(|x| x.to_owned()).collect()` boilerplate: `scope = scopes!("asd", "test")`.

Great idea! Probably I should make such a `scope!` PR firstly, and then merge this PR into it.

marioortizmanero:
We could just reuse a [macro from a crate like maplit](https://docs.rs/maplit/1.0.2/maplit/macro.hashset.html) until https://github.com/rust-lang/rfcs/issues/542 is worked on. Although [their implementation of `hashset!` isn't that complicated](https://github.com/bluss/maplit/blob/master/src/lib.rs#L77), but that way we could use stuff like `hashmap` in the endpoints, which could come in quite handy.

ramsayleung:
I just add a macro named `hashset!` in this commit: [`faf6543` (#189)](https://github.com/ramsayleung/rspotify/pull/189/commits/faf6543e989d92b4c1c88150ba8b5739cdb3cb41) instead of an additional PR. You could take time to review it :)

marioortizmanero:
Great! But I meant a specific macro for scopes, that even includes the `.to_owned()` part, and public for everyone to use. It could be a simple wrapper over your `hashset` macro, but we don't need a hashset macro for anything else so it could be a single macro as well.

ramsayleung:
Yep, I have just replaced `hashset!` with Rspotify-scope specific macro `scope!` in this commit: [`f78abf5` (#189)](https://github.com/ramsayleung/rspotify/pull/189/commits/f78abf5149b4a41f9cc8a341d8c6b3d9f54eec51)
And usage of this macro will look like:

```rust
let scope: HashSet<String> = scope!("playlist-read-private", "playlist-read-collaborative");
let tok = TokenBuilder::default()
    .access_token("test-access_token")
    .expires_in(Duration::seconds(1))
    .expires_at(Utc::now())
    .scope(scope)
    .refresh_token("...")
    .build()
    .unwrap();
```

src/lib.rs
```diff
@@ -173,6 +173,38 @@ compile_error!(
 
 #[doc(hidden)]
```
marioortizmanero, line 174: don't forget to remove this and use `#[macro_export]` instead

src/http/mod.rs
```diff
@@ -232,3 +232,68 @@ impl Spotify {
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::client::SpotifyBuilder;
+    use crate::oauth2::TokenBuilder;
+    use crate::scope;
+    use chrono::prelude::*;
+    use chrono::Duration;
+    #[test]
```
marioortizmanero, line 244: a newline before this would make it look better

marioortizmanero:
Can you use the new `scope` macro in the examples and other parts of code as well? Just search `.scope(` in the codebase

src/lib.rs
ramsayleung: Why did you like to add this `#[doc(hidden)]` attribute before? And do you mean use `#[macro_export]` for this `macros` module?

marioortizmanero: The `#[doc(hidden)]` attribute was added to make the macros private. But if you want to make it public you have to use `#[macro_export]`. Not sure if the former was necessary back when j used it myself, but I think I had trouble hiding the macro so I just used that.

ramsayleung: I think it's acceptable to keep the `macros` private, it's unnecessary to make it public. I prefer to recover `#[doc(hidden)]`.

marioortizmanero: But the whole point of `scopes!` is to make it easier to use for everyone.

src/lib.rs
```diff
@@ -173,6 +173,38 @@ compile_error!(
...
+    macro_rules! scope {
+	($($key:expr),*) => {{
+	    let mut container = ::std::collections::HashSet::new();
+	    $(
+		let _ = container.insert($key.to_owned());
+	    )*
+	    container
+	}
+    };
+}
```
marioortizmanero, line 207: The indentation here is off

src/lib.rs
```diff
@@ -173,6 +173,38 @@ compile_error!(
...
+    ///     .refresh_token("...")
+    ///     .build()
+    ///     .unwrap();
+    /// ```
+    #[macro_export]
+    macro_rules! scope {
+	($($key:expr),*) => {{
+	    let mut container = ::std::collections::HashSet::new();
+	    $(
+		let _ = container.insert($key.to_owned());
```
marioortizmanero, line 202: Also, is the `let _ =` necessary?

ramsayleung: Yep, It's unnecessary.

src/lib.rs
marioortizmanero: Could you mention the existence of the `scope!` macro in the token `scope` member? Those who don't take a look at the examples won't know about it. Here https://github.com/ramsayleung/rspotify/blob/master/src/oauth2.rs#L88, something like "you can use the macro `scopes!` to build it at compile time easily".

ramsayleung: Do you think `macro` module is necessary, or we could just put the macro in the `lib.rs` as other open source projects like `serde`: https://docs.serde.rs/src/serde/macros.rs.html#112-120

marioortizmanero: `serde` is also inside a module, because they are in the `macros.rs` file. But if you consider it cleaner that way, then go ahead.

ramsayleung: I think `macros.rs` is a good idea, and I will move `json_insert` and `scope` into `macros.rs`

marioortizmanero:
We can continue with this PR after merging the ID's PR. As to my review, the only missing things are:

* > Could you mention the existence of the `scope!` macro in the token `scope` member? Those who don't take a look at the examples won't know about it. Here https://github.com/ramsayleung/rspotify/blob/master/src/oauth2.rs#L88, something like "you can use the macro `scopes!` to build it at compile time easily".
  > 
  > Also now that I see it, `#[builder(default = "HashSet::new()")]` isn't needed, you can just change it to `#[builder(default)]`

* Lines 30-31 in `macros.rs` should be merged in the same one with the same indentation to keep it consistent (for some reason `cargo fmt` doesn't format macros automatically).

* And the merge conflicts.

* Also, should the name be `scope!` or `scopes!`? I actually prefer the latter because it builds a set of scopes, not a single one. But not sure if you named it `scope` for a different reason

ramsayleung:
> Could you mention the existence of the scope! macro in the token scope member? Those who don't take a look at the examples won't know about it. Here https://github.com/ramsayleung/rspotify/blob/master/src/oauth2.rs#L88, something like "you can use the macro scopes! to build it at compile time easily".

Great idea.

> Lines 30-31 in macros.rs should be merged in the same one with the same indentation to keep it consistent (for some reason cargo fmt doesn't format macros automatically)

I have the same problem that `cargo fmt` doesn't work for `macros` for unknown reason, so I have to format it manually

> Also, should the name be scope! or scopes!? I actually prefer the latter because it builds a set of scopes, not a single one. But not sure if you named it scope for a different reason

Great catch. No special reason, I just forgot to take context into consideration, so simply came with this name `scope!`.

ramsayleung:
> This PR can be merged after these suggestions, right?

Yep, this PR is ready to be merged :-)

PR #191, opened 2021-03-07
Move model into a separate rspotify-model crate

udoprog:
## Description

This extracts the models used into a separate crate, so they can be used separately from the client implementation.

## Motivation and Context

I have a separate client which is not suitable for general use that does not need the client bits, but could benefit greatly from the models.

## Type of change

This also requires that the next release of `rspotify` releases the `rspotify-model` crate. The separated crate is being re-exported with the same API as before. A few models have been modified from being `pub(in crate)` to just be `pub`.

## How Has This Been Tested?

Just built for now. I don't have the necessary plumbing to run the integration tests, so we'll see if they pass through Github Actions.


marioortizmanero:
I like the idea. Other libraries are much more modular than `rspotify`, see [bevy](https://crates.io/search?q=bevy).

But shouldn't we use workspaces rather than a crate in `src` and another in `model`? Maybe it'll be more organized that way

udoprog:
> But shouldn't we use workspaces rather than a crate in `src` and another in `model`? Maybe it'll be more organized that way

Yeah. Just making the change as small as possible right now. Setting up a workspace would mean moving rspotify proper into its own directory which I'm open to doing once this has gotten a green light.

marioortizmanero:
Sounds good to me, we just need to know @ramsayleung's opinion. We could move some parts of the code into crates, like the HTTP client as `rspotify-http`, but not sure that's worth our time.

The only part I can see getting worse is the documentation. We would have to add a link to `rspotify-model`'s crate on the main docs or something.

udoprog:
> The only part I can see getting worse is the documentation. We would have to add a link to `rspotify-model`'s crate on the main docs or something. It might be harder to upload new versions and such, but I guess that's nothing a good GitHub Actions workflow can't fix.

With the `rspotify` crate we can mark the re-export with `#[doc(inline)]`. Then rustdoc will pretend that whatever types are being re-exported are part of the same crate, so the experience should be the same.

> Also, if you want to fix the CI you should merge with master.

Will do, thanks!

marioortizmanero:
> With the rspotify crate we can mark the re-export with #[doc(inline)]. Then rustdoc will pretend that whatever types are being re-exported are part of the same crate, so the experience should be the same.

Oh that is great, thanks. We have some models that are supposedly private (those with `pub (in crate)` previously). To keep them that way we could use `#[doc(hidden)]`.

Also, you answered just after I edited my comment, I think you missed this:

> By the way, may I know why you need the model but not the client? Not sure what your use case is.

udoprog:
> By the way, may I know why you need the model but not the client? Not sure what your use case is.

Sure. I already have a client that is integrated into the project <https://github.com/udoprog/OxidizeBot/blob/main/bot/src/api/spotify/mod.rs>. It uses an 17 month old copy of the models provided in rspotify.

marioortizmanero:
> Sure. I already have a client that is integrated into the project https://github.com/udoprog/OxidizeBot/blob/main/bot/src/api/spotify/mod.rs. It uses an 17 month old copy of the models provided in rspotify.

Oh ok that makes sense. Why not just use `rspotify` entirely in this case though? You'd save some maintainance work that way. Anyhow, I like the idea, thanks for contributing.

ramsayleung:
It's a good idea to seperate rspotify-model :) I agree with Mario's point:

> But shouldn't we use workspaces rather than a crate in `src` and another in `model`?

`src` and `model` don't stand on the same level of abstraction, IMHO, it will be better to use workspace.

> We could move some parts of the code into crates, like the HTTP client as `rspotify-http`, but not sure that's worth our time.

Or we could refactor the HTTP client into `rspotify-http` crate and move authenticate process into `rspotify-auth`? I am not sure about that either.

Since the architecture of `rspotify` package would look like this:

image

We could seperate `rspotify` into crates like this:

+ `rspotify-model`
+ `rspotify-oauth`
+ `rspotify-http`
+ `rspotify-client`

or 

+ `rspotify-model`
+ `rspotify-http`
+ `rspotify-client`

How about this idea?

udoprog:
> `src` and `model` don't stand on the same level of abstraction, IMHO, it will be better to use workspace.

As indicated earlier, I don't have anything against doing the move as long as thing change is approved. But since I won't be using either the client, auth, or http I'm not the best suited to moving the other crates.

I can try and introduce a workspace and do `rspotify-model` though if you want. But as indicated earlier, it will be a lot of changes and they'll make any open PRs outdated. I'd also stress that in principle, there's nothing preventing keeping things outside of a workspace structure in the interim.

marioortizmanero:
Any progress on this, @udoprog?

udoprog:
> Any progress on this, @udoprog?

No. Anyone else interested in doing the split can feel free to!

marioortizmanero:
Sure, I'll give it a try this weekend if I have time.

ramsayleung:
Are you guys working on split this crate into workspace? If not, I could take time to do that :)

marioortizmanero:
Hi! Yes, I tried it last week, but failed to, in this branch https://github.com/ramsayleung/rspotify/tree/separate-crates. I split up the library into `rspotify-http`, `rspotify-auth` and `rspotify-client`. But that ended up being impossible because they are tied together and don't work in separate crates at all until https://github.com/ramsayleung/rspotify/issues/173 is done.

So we should only have `rspotify-client` with these three crates combined, and after #173 we can separate them again.

I was going to try to do that but I didn't end up having enough time. You can give it a try if you want, as the branch is still there and you can make a new one based on `separate-crates`. I was planning on doing that myself this weekend but not sure if I'll be able to either.

Sorry about that!

ramsayleung:
Since you are working on separating this library into individual crates, we could wait util #173 is merged :)

marioortizmanero:
This has been superseded by #199, so closing the PR.

PR #194, opened 2021-03-09
Allow `PlaylistItem` to contain episodes.

kstep:
## Description

It's a change for meta-issue #127: "Allow `PlaylistTrack`/`PlaylistItem` to contain episodes."

## Motivation and Context

According to meta-issue #127 and [official docs](https://developer.spotify.com/documentation/web-api/reference/#object-playlisttrackobject), `PlaylistItem` object can have either `FullTrack` or `FullEpisode` object as a value of the `track` field.
Hence the fix.

## Dependencies 

Update docs and changelog.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

Run a full test suite.

---
This change is [Reviewable](https://reviewable.io/reviews/ramsayleung/rspotify/194)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ If we missed any change or there's something you'd like to discuss about this ve
   + `Device`
   + `DevicePayload`
   + `Image`
-  + `PlayingItem` 
+  + `PlayableItem`
   + `Offset`
   + `Page`
   + `CursorBasedPage`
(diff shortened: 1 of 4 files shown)


kstep:
As a first version, I just changed the `PlaylistItem::track` field type to `Option<PlayingItem>`.
It's not the best solution, as it renders the `PlayingItem` type name imperfect for its function. It should be either renamed (please advise on the better name), or a new type should be introduced. I don't like to introduction of a new type, as it will be just a duplicate of the `PlayingItem` type, which is not DRY at all.

marioortizmanero:
Why is `PlayingItem`'s name imperfect? Both a track and an episode are playable types.

kstep:
Because if the item is part of `PlaylistItem`, it is not necessary playing at this point in time. If you think it's OK, then we can go on with this PR.

kstep:
Renamed `PlayingItem` to `PlayableItem` and updated the changelog.

marioortizmanero (approved):

marioortizmanero:
Merged and updated the meta-issue.

PR #198, opened 2021-04-09
Added missing parameter from current_user_recently_played()

hellbound22:
## Description

Fix the implementation of ```current_user_recently_played()``` with the missing parameters.

## Motivation and Context

The Spotify web api of ```current_user_recently_played()``` has the parameters ```before``` and ```after``` that are not implemented in this crate.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

The OAuth2 test passes when adding the ```None``` as the second parameter. Might have to write another test when the Option is ```Some(_)```

Also, been using this with success in a project of mine.

- [x] ```test_with_oauth::test_current_user_recently_played()```


hellbound22:
We could change the type of the values of the new TimeLimits enum from i64 (representing the actual timestamp) to Utc and then convert to the timestamp equivalent in before the request itself

src/client.rs
```diff
@@ -1079,15 +1079,28 @@ impl Spotify {
...
+    pub async fn current_user_recently_played<L: Into<Option<u32>>, T: Into<Option<TimeLimits>>>(
         &self,
         limit: L,
+        time_limit: T,
     ) -> ClientResult<CursorBasedPage<PlayHistory>> {
-        let mut params = Query::with_capacity(1);
+        let mut params = Query::with_capacity(2);
         params.insert("limit".to_owned(), limit.into().unwrap_or(50).to_string());
+        if let Some(l) = time_limit.into() {
+            match l {
```
ramsayleung, line 1094: Perhaps you could give these variables a more descriptive name than `l` and `y`, it's a little bit meaningless.

src/client.rs
```diff
@@ -1079,15 +1079,28 @@ impl Spotify {
...
+                }
+                TimeLimits::After(y) => {
+                    params.insert("after".to_owned(), y.to_string());
+                }
+            };
+        }
+
         let result = self
             .endpoint_get("me/player/recently-played", &params)
             .await?;
```
ramsayleung, line 1106: I suppose that you are using Vim to edit files, since you are commiting the swap file `src/model/enums/.misc.rs.swp ` to this repository. It's useful for you, but not for us :) (just a joke)

src/model/enums/misc.rs
```diff
@@ -36,6 +36,15 @@ pub enum TimeRange {
     ShortTerm,
 }
 
+/// Time limits in miliseconds (unix timestamps)
+///
+/// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-the-users-recently-played-tracks/)
```
ramsayleung, line 41: The reference seems to wrong, it refers to the Home page.

PR #199, opened 2021-04-13
Separate crates

marioortizmanero:
## Description

Some work towards #191. I've ended up separating the crates but leaving the auth, http and client ones mixed under `rspotify_client` until #173 is worked on.

## Motivation and Context

See #191

## Dependencies 

Should be the same

## Type of change

This change means that types like `FullAlbum` in the model now have to be public.

## How Has This Been Tested?

The previous tests still work now. Same goes for the examples (which by the way I had to fix because they weren't working)

---

Closes #191


marioortizmanero:
I think the best organization for our crates will be the following:

* `src/`: the Spotify client, with the endpoints and everything needed to access the API
* `rspotify-model/`: with the model
* `rspotify-http/`: with the HTTP client abstraction features
* `rspotify-auth/`: with the Auth functionalities, exclusively

In this PR `rspotify-http`, `rspotify-auth` and `src` are inside `src` because they are too tied together. I was going to make the client a separate crate, like `rspotify-client` but figured it's not really worth it. I can understand that someone might want to use `rspotify-model ` or `rspotify-http` on their own, but using `rspotify-client` would be just like using `rspotify`, which is just confusing, in my opinion.

marioortizmanero:
This is now ready for review. Please @ramsayleung take a look at #166 after merging this; it's been open for a while and I would like to finish it before #173 if possible.

ramsayleung:
Since `rspotify-http`, `rspotify-auth` are too tied together, and we have to keep working on the separating crates after #173 is done, would you like to consider working on this PR after #173 and #166 merged, so we don't have to change it for twice, just to reduce some unnecessary context-switch work

> Please @ramsayleung take a look at #166 after merging this; it's been open for a while and I would like to finish it before #173 if possible.

Sorry for this, I don't realize that it's ready to review, I will check it on the weekend :)

marioortizmanero:
> Since rspotify-http, rspotify-auth are too tied together, and we have to keep working on the separating crates after #173 is done, would you like to consider working on this PR after #173 and #166 merged, so we don't have to change it for twice, just to reduce some unnecessary context-switch work

The thing is that this is already discussed and implemented, so it's preferable to do it in two steps in my opinion. It doesn't really matter that it's done all at once.

> Sorry for this, I don't realize that it's ready to review, I will check it on the weekend :)

Great, thanks!

marioortizmanero:
If you'd rather not merge this, I can just start #173 on a new PR based on this branch instead of master, by the way.

ramsayleung:
> The thing is that this is already discussed and implemented, so it's preferable to do it in two steps in my opinion. It doesn't really matter that it's done all at once. This way that the #173 PR will be split up and probably easier to review, and I have more clear how the structure is supposed to go, too.

It will be easier to review if you split it into two PR, and polishing it step by step is also a good idea. I would merge this after I review it. But before that, we need to keep our eyes on #166 :-)

marioortizmanero:
This will be marked as a draft until I implement #173, and I'll build said PR over this branch. That way the work will be split up into various PRs instead of a huge one.

ramsayleung:
I think the separated crates would look like this:

rspotify-arch

ramsayleung (approved):

ramsayleung:
I am not sure if this PR ready to merge, even though I have merged all separated PRs

marioortizmanero:
Yes, you can merge it.

ramsayleung:
Finally, it takes so long, but we reached it :)

marioortizmanero:
Yes I can't believe it! We're getting closer to v0.11 :)

PR #200, opened 2021-04-19
Replace HashMap<String, String> with HashMap<&str, &str> to reduce allocation

### LINKED ISSUE #127: Meta-Issue

ramsayleung:
## Description

Try to fix #127 and #160 , use `HashMap<&str, &str>` instead of `HashMap<String, String>` to avoid so many `to_string` and `to_owned`.

## Motivation and Context

Reduce uncessary allocation.

## Type of change

## How Has This Been Tested?

All existing tests pass


Cargo.toml
```diff
@@ -110,3 +110,6 @@ path = "examples/ureq/search.rs"
 name = "seek_track"
 required-features = ["env-file", "cli", "client-ureq"]
 path = "examples/ureq/seek_track.rs"
+
+[profile.dev]
+split-debuginfo = "unpacked"
```
marioortizmanero, line 115: is this to get faster compilation times on MacOS? It'll become the default in the  next version (I think) https://github.com/rust-lang/cargo/pull/9298 so you should add a note so that we can remove it in the future

src/model/enums/country.rs
```diff
@@ -1,11 +1,11 @@
 use serde::{Deserialize, Serialize};
-use strum::ToString;
+use strum::{AsRefStr, ToString};
 
 /// ISO 3166-1 alpha-2 country code, from
 /// [country-list](https://datahub.io/core/country-list)
 ///
 /// [Reference](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2)
-#[derive(Clone, Serialize, Deserialize, Copy, PartialEq, Eq, Debug, ToString)]
+#[derive(Clone, Serialize, Deserialize, Copy, PartialEq, Eq, Debug, ToString, AsRefStr)]
```
marioortizmanero, line 8: If we're using `AsRefStr` we might as well remove the `ToString`, right? (this happens in many other places)

marioortizmanero (commented):
Yes, I really wanted this! Great job.

Cargo.toml
ramsayleung: Removed :)

src/model/enums/country.rs
ramsayleung: Removed :)

marioortizmanero (approved):

PR #202, opened 2021-04-19
Remove Rspotify default parameters and add parameter macros

### LINKED ISSUE #134: Optional parameters

marioortizmanero:
## Description

* Leaves default parameter handling to Spotify, removing `.unwrap_or`
* Implements a macro for both query and json parameters so that endpoints are much nicer to read.
* Closes #134, we're consistently using `Option<T>` instead of `Into<Opton<T>>` now. If you don't like this please let me know, but I didn't get any more feedback so I went for the simpler option.

## Motivation and Context

* Default values should be assigned by the Spotify API instead by us, clearly. We shouldn't touch things we don't have to.
* As #127 said, endpoints should be as concise as possible because they're the most important part of the library after all, and it's important that they are easy to maintain.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

- [x] There are a couple new tests for the macros

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -621,9 +621,9 @@ impl Spotify {
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
             opt fields,
+            opt market => market.as_ref(),
             opt limit => &limit,
             opt offset => &offset,
-            opt market => market.as_ref(),
         };
 
         let url = format!("playlists/{}/tracks", playlist_id.id());
@@ -745,10 +745,10 @@ impl Spotify {
         track_ids: impl IntoIterator<Item = &'a TrackId>,
     ) -> ClientResult<()> {
         let uris = track_ids.into_iter().map(|id| id.uri()).collect::<Vec<_>>();
-
         let params = build_json! {
-            req uris => uris
+            req uris
         };
+
         let url = format!("playlists/{}/tracks", playlist_id.id());
         self.endpoint_put(&url, &params).await?;
 
@@ -1063,8 +1063,8 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let params = build_map! {
             req r#type => Type::Artist.as_ref(),
-            opt after => &after,
-            opt limit => &limit,
+            opt after,
+            opt limit,
         };
 
         let result = self.endpoint_get("me/following", &params).await?;
@@ -1447,9 +1447,9 @@ impl Spotify {
         let limit = limit.map(|x| x.to_string());
         let offset = offset.map(|x| x.to_string());
         let params = build_map! {
+            opt market => market.as_ref(),
             opt limit => &limit,
             opt offset => &offset,
-            opt market => market.as_ref(),
         };
 
         let result = self.endpoint_get("browse/new-releases", &params).await?;
@@ -1809,7 +1809,7 @@ impl Spotify {
                 Offset::Position(position) => json!({ "position": position }),
                 Offset::Uri(uri) => json!({ "uri": uri.uri() }),
             },
-            opt position_ms => position_ms
+            opt position_ms,
 
         };
 
@@ -2120,9 +2120,9 @@ impl Spotify {
         let limit = limit.map(|x| x.to_string());
         let offset = offset.map(|x| x.to_string());
         let params = build_map! {
+            opt market => market.as_ref(),
             opt limit => &limit,
             opt offset => &offset,
-            opt market => market.as_ref(),
         };
 
         let url = format!("shows/{}/episodes", id.id());
@@ -2191,7 +2191,7 @@ impl Spotify {
     ) -> ClientResult<Vec<bool>> {
         let ids = join_ids(ids);
         let params = build_map! {
-            req ids => ids.as_str(),
+            req ids => &ids,
         };
         let result = self.endpoint_get("me/shows/contains", &params).await?;
         self.convert_result(&result)
--- a/src/client.rs
+++ b/src/client.rs
@@ -721,7 +721,7 @@ impl Spotify {
         let uris = track_ids.into_iter().map(|id| id.uri()).collect::<Vec<_>>();
         let params = build_json! {
             "uris": uris,
-            "position": position,
+            optional "position": position,
         };
 
         let url = format!("playlists/{}/tracks", playlist_id.id());
--- a/src/client.rs
+++ b/src/client.rs
@@ -194,7 +194,7 @@ impl Spotify {
     ) -> ClientResult<Vec<FullTrack>> {
         let ids = join_ids(track_ids);
         let params = build_map! {
-            opt market => market.as_ref(),
+            optional market => market.as_ref(),
         };
 
         let url = format!("tracks/?ids={}", ids);
@@ -274,10 +274,10 @@ impl Spotify {
         let limit = limit.map(|x| x.to_string());
         let offset = offset.map(|x| x.to_string());
         let params = build_map! {
-            opt album_type => album_type.as_ref(),
-            opt market => market.as_ref(),
-            opt limit,
-            opt offset,
+            optional album_type => album_type.as_ref(),
+            optional market => market.as_ref(),
+            optional limit,
+            optional offset,
         };
 
         let url = format!("artists/{}/albums", artist_id.id());
@@ -300,7 +300,7 @@ impl Spotify {
         market: Market,
     ) -> ClientResult<Vec<FullTrack>> {
         let params = build_map! {
-            req market => market.as_ref()
+            required market => market.as_ref()
         };
 
         let url = format!("artists/{}/top-tracks", artist_id.id());
@@ -386,12 +386,12 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
-            req q,
-            req r#type => r#type.as_ref(),
-            opt market => market.as_ref(),
-            opt include_external => include_external.as_ref(),
-            opt limit,
-            opt offset,
+            required q,
+            required r#type => r#type.as_ref(),
+            optional market => market.as_ref(),
+            optional include_external => include_external.as_ref(),
+            optional limit,
+            optional offset,
         };
         println!("params: {:#?}", params);
 
@@ -431,8 +431,8 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
-            opt limit,
-            opt offset,
+            optional limit,
+            optional offset,
         };
 
         let url = format!("albums/{}/tracks", album_id.id());
@@ -468,8 +468,8 @@ impl Spotify {
         market: Option<Market>,
     ) -> ClientResult<FullPlaylist> {
         let params = build_map! {
-            opt fields,
-            opt market => market.as_ref(),
+            optional fields,
+            optional market => market.as_ref(),
         };
 
         let url = format!("playlists/{}", playlist_id.id());
@@ -504,8 +504,8 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
-            opt limit,
-            opt offset,
+            optional limit,
+            optional offset,
         };
 
         let result = self.endpoint_get("me/playlists", &params).await?;
@@ -544,8 +544,8 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
-            opt limit,
-            opt offset,
+            optional limit,
+            optional offset,
         };
 
         let url = format!("users/{}/playlists", user_id.id());
@@ -569,7 +569,7 @@ impl Spotify {
         fields: Option<&str>,
     ) -> ClientResult<FullPlaylist> {
         let params = build_map! {
-            opt fields,
+            optional fields,
         };
 
         let url = match playlist_id {
@@ -620,10 +620,10 @@ impl Spotify {
         let limit = limit.map(|s| s.to_string());
         let offset = offset.map(|s| s.to_string());
         let params = build_map! {
-            opt fields,
-            opt market => market.as_ref(),
-            opt limit,
-            opt offset,
+            optional fields,
+            optional market => market.as_ref(),
+            optional limit,
+            optional offset,
         };
 
         let url = format!("playlists/{}/tracks", playlist_id.id());
@@ -651,10 +651,10 @@ impl Spotify {
         description: Option<&str>,
     ) -> ClientResult<FullPlaylist> {
         let params = build_json! {
-            req name,
-            opt public,
-            opt collaborative,
-            opt description,
+            required name,
+            optional public,
+            optional collaborative,
+            optional description,
         };
 
         let url = format!("users/{}/playlists", user_id.id());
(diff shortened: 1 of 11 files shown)


ramsayleung:
Great job! I totally agree to your point, the default parameter in this crate is a little messy. Good to see the improvement.

Since the parameters in endpoints are mostly passed by value, then we are working on the function signature, shall we update the parameters to pass them by reference?

marioortizmanero:
> Since the parameters in endpoints are mostly passed by value, then we are working on the function signature, shall we update the parameters to pass them by reference?

I don't really know. For now I've only set those parameters that I needed by reference, and left the rest by value. I guess it would make sense, yeah.

This PR is now ready for review when you have time, by the way. It ended up much bigger than I thought it would haha. Let me know if you have any doubts/suggestions regarding the macros.

ramsayleung:
> For now I've only set those parameters that I needed by reference, and left the rest by value.

I think it will make the endpoint inconsistency, some parameters passed by reference, and the others by value. I am not sure if it's a chance to modify them all.

> This PR is now ready for review when you have time, by the way.

Yes, I will.

marioortizmanero:
> I think it will make the endpoint inconsistency, some parameters passed by reference, and the others by value. I am not sure if it's a chance to modify them all.

I'd say go for it, it makes sense to me.

src/macros.rs
```diff
@@ -33,38 +33,209 @@ macro_rules! scopes {
...
+        let mut params = $crate::http::Query::with_capacity(
+            $crate::count_items!($( $name ),*)
+        );
+        $(
+            $crate::params_internal!(
+                params,
+                $kind,
+                $name,
+                $crate::ident_str!($name),
+                $crate::opt!($( $val )?, $name)
```
ramsayleung, line 130: Ooooh, this macro is a little complex, I take times to understand it, and the key point is `$( => $val:expr )?`, the `?` means "at most one" repetition.

if `$( => $val:expr )` doesn't repeat, then the `$crate::opt!($( $val )?, $name)` will match `(, $default:expr)` arm, set the `$name` as `$val`

otherwise it will match `($optional:expr, $default:expr)`, set the `$( $val )` as `$val`.

src/macros.rs
```diff
@@ -33,38 +33,209 @@ macro_rules! scopes {
...
+#[doc(hidden)]
+#[macro_export]
+macro_rules! count_items {
+    ($($item:expr),*) => {<[()]>::len(&[$($crate::replace_expr!($item ())),*])};
+}
+
+/// Hack to convert an identifier to a string, including raw identifiers.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! ident_str {
```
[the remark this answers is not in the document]
marioortizmanero: It's for raw *identifiers* like `r#type`, not raw *strings* like `r#"type"#`

src/macros.rs
marioortizmanero: Yup, that's it. The thing is that if ` $( => $val:expr )?` doesn't repeat is output is empty, and I needed a "default" value.

src/macros.rs
```diff
@@ -33,38 +33,209 @@ macro_rules! scopes {
...
+        }
+    }};
+}
+
+/// Private macro to insert either required or optional fields. Pattern matching
+/// will accordingly pick the branch, and then insert ($key, $val) into $map.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! params_internal {
+    ($map:ident, req, $name:ident, $key:expr, $val:expr) => {
```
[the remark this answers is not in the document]
marioortizmanero: I don't really like that we have to write the entire word, as the macro is meant to have less code, but yes, I think it'll be easier to understand for a newcomer to the library, so it might be worth it. I've just updated it, let me know what you think.

marioortizmanero:
Something that's inconsistent about the `build_map` macro that I'm unsure about is that for `optional` parameters it takes a reference, and for `required` it's the value itself. This is because for `optional` parameters it may be expanded to:

```rust
if let Some(ref name) = name {
     params.insert("name", name);
}
```

But for required params it's:

```rust
params.insert("name", name);
``` 

Which doesn't need a reference. The effect of this is that you can pass a String to an `optional` parameter and it will be converted to `&str` automatically, but this won't work for `required` parameters. What do you think? Perhaps it should use `if let Some(name) = name` instead?

marioortizmanero:
Oooh or another idea for the macro syntax, which solves the previous issue. What if it worked like this instead, with pattern matching?

```rust
let params = build_map! {
    name, // this is required
    Some(ref limit), // this is optional
    Some(ref market) => market.as_ref() // you can still add a specific value
};
```

marioortizmanero:
The previous syntax would technically allow to turn this:

```rust
let params = build_json! {
    required uris => uris.iter().map(|id| id.uri()).collect::<Vec<_>>(),
    optional position_ms,
    optional offset => match offset {
        Offset::Position(position) => json!({ "position": position }),
        Offset::Uri(uri) => json!({ "uri": uri.uri() }),
    },
};
```

into this:

```rust
let params = build_json! {
    uris => uris.iter().map(|id| id.uri()).collect::<Vec<_>>(),
    position_ms,
    Offset::Position(offset) => json!({ "position": position }),
    Offset::Uri(offset) => json!({ "uri": uri.uri() }),
};
```

You could still use the first approach, but can now use pattern matching and turn it into the second one. But it might be more confusing, idk.

marioortizmanero:
I think that the best macro syntax is the last one I mentioned, because I dislike magic macros that make you look their documentation in order to understand them. And in this case it's pretty much self-descriptive because it doesn't really hide anything, unlike the other two macros.

What do you think? Any other macro you can come up with?

Sorry for thinking about other syntaxes *after* imlementing the macro, I didn't really notice the downsides of the current one until it was fully implemented.

ramsayleung:
> What do you think? Any other macro you can come up with?

I personally prefer the last one with same reason as you do. The other macro syntaxes are a little subtle and hide something into a black box. If you want to know about how it works, you have dig deep into it.

```rust
let params = build_map! {
    required "name": name, // will always be inserted
    "name": name, // we could remove the `required` keyword
    optional "limit": limit,
    optional "market": market.map(|m| m.as_ref()), // this is like the  `=>` expressions
};
```

And I think we could remove the `required` keyword, since we have explicitly used the `optional` keyword, on the opposite, the `required` keyword is not that necessary.

src/macros.rs
```diff
@@ -33,38 +33,210 @@ macro_rules! scopes {
...
+        }
+    }};
+}
+
+/// Private macro to insert either required or optional fields. Pattern matching
+/// will accordingly pick the branch, and then insert ($key, $val) into $map.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! params_internal {
+    ($map:ident, required, $name:ident, $key:expr, $val:expr) => {
```
ramsayleung, line 84: As what I commented before, we could remove the `required` keyword as the parameters are required by default if we don't add the `optional` keyword.

src/macros.rs
```diff
@@ -33,38 +33,210 @@ macro_rules! scopes {
...
+        }
+    };
+}
+
+/// This macro and [`build_json`] help make the endpoints as concise as possible
+/// and boilerplate-free, which is specially important when initializing the
+/// parameters of the query with a HashMap/similar. Their items follow the
+/// syntax:
+///
+///   (required|optional) key [=> value]
```
ramsayleung, line 100: Shall we add an example for this syntax? it looks a little foreign. 

```
/// For more information, please refer to the `test::test_build_map` function in
/// this module to see an example, or the real usages in Rspotify's client.
```

Or just move this description up

src/macros.rs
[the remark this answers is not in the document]
ramsayleung: Yup, I get it now

src/macros.rs
```diff
@@ -33,38 +33,210 @@ macro_rules! scopes {
...
+/// will accordingly pick the branch, and then insert ($key, $val) into $map.
+#[doc(hidden)]
+#[macro_export]
+macro_rules! params_internal {
+    ($map:ident, required, $name:ident, $key:expr, $val:expr) => {
+        $map.insert($key, $val);
+    };
+    ($map:ident, optional, $name:ident, $key:expr, $val:expr) => {
+        // Will only insert when `$name` is not None.
+        if let Some(ref $name) = $name {
```
ramsayleung, line 89: If this statement will only insert when `$name` is not None, would you like to replace `if let Some(ref $name) = $name ` with `if $name.is_some()`. I was confused with this statement in the beginning, it get the reference of `$name`, and what's the next step. I don't get it unill I noticed the comment.

src/client.rs
```diff
@@ -382,28 +374,26 @@ impl Spotify {
...
+        let offset = offset.map(|s| s.to_string());
+        let params = build_map! {
+            required q,
+            required r#type => r#type.as_ref(),
+            optional market => market.as_ref(),
+            optional include_external => include_external.as_ref(),
+            optional limit,
+            optional offset,
+        };
+        println!("params: {:#?}", params);
```
ramsayleung, line 396: We could remove this debug statement.

marioortizmanero:
Uggghhhh I've implemented the last macro I suggested, and while the macro's code is much simpler (I can remove the whole `opt` and `ident_str` macros), It still has some inconsistency issues...

Let me show you an example. In Rust, you can't do this:

```rust
let x: Option<String> = Some("abc".to_string());
let y: Option<&str> = x.as_ref(); // This is Option<&String>, so it will fail
```

But for some reason, with the new macro, this will convert  `limit` from a `Option<String>` into a `&str`:

```rust
let params = build_map! {
    optional "limit": limit.as_ref() // No compilation error here??
};
```

Which doesn't make much sense, because the hashmap insertion requires an `Option<&str>` and we have a  `Option<&String>`, right?

Well, turns out that if you expand the macro, you get this:

```rust
if let Some(val) = limit.as_ref() {
    // val is a &String here, which is what we expected.

    // But now when we insert the value, deref coercion converts it into a &str
    params.insert(key, val);

    // Meaning that the previous line gets converted to:
    //   params.insert(key, &*val);
    // Where the inserted value *is* a &str.
}
```

The required parameter gets converted to this, simply:

```rust
params.insert(key, val.as_ref());
```

Where deref coercion doesn't take place...

This ultimately means that:

```rust
let params = build_map! {
    "limit": limit.as_ref(), // this will fail
    optional "limit": limit.as_ref(), // this will work
};
```

Which is just weird and confusing, in my opinion.

---

Just letting you know about what I'm dealing with right now and why it might take me more to implement this than I expected. If you have any ideas for a nice & consistent macro do let me know.

src/client.rs
marioortizmanero: done

marioortizmanero:
So, do you think the endpoints look better like this? In that case I think this is ready for a merge.

src/client.rs
```diff
@@ -777,30 +758,33 @@ impl Spotify {
...
-            "range_start": range_start,
-            "range_length": range_length.into().unwrap_or(1),
-            "insert_before": insert_before
-        });
-        if let Some(snapshot_id) = snapshot_id {
-            json_insert!(params, "snapshot_id", snapshot_id);
-        }
+        let uris = uris.map(|u| u.iter().map(|id| id.uri()).collect::<Vec<_>>());
+        let params = build_json! {
+            "playlist_id": playlist_id,
```
ramsayleung, line 781: Is it a mistake that the `playlist_id` is a part of the JSON body parameter?

src/client.rs
```diff
@@ -1040,48 +1022,49 @@ impl Spotify {
...
-        let mut params = Query::with_capacity(2);
-        let limit = limit.into().unwrap_or(20).to_string();
-        params.insert("limit", &limit);
-        params.insert("type", Type::Artist.as_ref());
-        if let Some(ref after) = after {
-            params.insert("after", &after);
-        }
+        let limit = limit.map(|s| s.to_string());
+        let params = build_map! {
+            "r#type": Type::Artist.as_ref(),
```
ramsayleung, line 1064: It seems we haven't handle the raw identifiers, then the `r#type` will be inserted into the parameter map?

src/client.rs
```diff
@@ -1593,18 +1570,27 @@ impl Spotify {
...
         &self,
+        payload: &Map<String, Value>,
         seed_artists: Option<Vec<&ArtistId>>,
-        seed_genres: Option<Vec<String>>,
+        seed_genres: Option<Vec<&str>>,
         seed_tracks: Option<Vec<&TrackId>>,
-        limit: L,
-        market: Option<Market>,
-        payload: &Map<String, Value>,
+        limit: Option<u32>,
```
ramsayleung, line 1579: The `limit` parameter should be the last parameter as what `CHANGELOG.md` claims.

src/client.rs
```diff
@@ -1802,20 +1763,17 @@ impl Spotify {
...
-            "me/player",
-            &json! ({
-                "device_ids": vec![device_id.to_owned()],
-                "play": force_play.into().unwrap_or(true)
-            }),
-        )
-        .await?;
+        let params = build_json! {
+            "device_ids": [device_id],
+            optional "force_play": force_play,
```
ramsayleung, line 1773: The field name should be `play` instead of `force_play`.

src/client.rs
marioortizmanero: whoops, my bad

src/client.rs
marioortizmanero: yup, removed

src/client.rs
marioortizmanero: fixed

src/client.rs
marioortizmanero: good catch

src/client.rs
```diff
@@ -740,11 +719,11 @@ impl Spotify {
         position: Option<i32>,
     ) -> ClientResult<PlaylistResult> {
         let uris = track_ids.into_iter().map(|id| id.uri()).collect::<Vec<_>>();
+        let params = build_json! {
+            "uris": uris,
+            "position": position,
```
ramsayleung, line 724: The `position` should be optional, right?

marioortizmanero: Fixed

ramsayleung (approved):

ramsayleung:
Merged

> So, do you think the endpoints look better like this? In that case I think this is ready for a merge.

Yep, both macros and endpoints look better than before :)

PR #207, opened 2021-04-29
[WIP] Auth restructure

### LINKED ISSUE #173: Restructure the authentication process

### LINKED ISSUE #208: current_user_playing_track return 401
**Describe the bug**
the method in the spotify client `current_user_playing_track` return `Err( StatusCode(401, "Unauthorized", ),)`

**To Reproduce**
1. Build oauth
2. Build spotify
3. Add or set the scope to `user-read-currently-playing`
4. prompt user for token.
5. call the method `current_user_playing_track`

**Expected behavior**
An struct of CurrentlyPlayingContext

**Log/Output data**
```
Err(
    StatusCode(
        401,
        "Unauthorized",
    ),
)
```

**Additional context**
To solve the problem in the method `current_user_playing_track` change `self.get` to `self.endpoint_get`

marioortizmanero:
TODO:

- [x] Implement everything listed in #173 for the new interface. This also allows us to remove `derive_builder` completely from the library.
- [x] Split the http module into a different crate
- [ ] Implement PKCE authentication (maybe for another PR?)
- [x] Export some parts of the library as top-level imports (requested in #127, but we might want to add more?)
- [x] Declare a common `Config` struct with things like `token_cached` or `token_refreshing` for #135. The former is implemented here, but the latter is left as a TODO.
- [x] Fix the examples
- [x] Rename tests like `album.rs` into `client_credentials.rs`, teaching about the authentication process instead.
- [x] Fix the tests
- [x] Make sure the endpoints are distributed properly into the different clients, and same for the tests.
- [ ] Tidy up and self-review
- [ ] Try to keep as few things from the Rspotify clients as possible in `rspotify-http`
- [ ] Update changelog
- [x] Update documentation
- [x] Make sure the feature collision feature still works

Note that this only really separates the `rspotify-http` crate, as I don't think moving the auth logic to a separate crate is really worth it. We can discuss it but I'm not sure it's necessary at all.

---

Closes #207  
Closes #208


marioortizmanero:
Ok so turns out this trait-based implementation is only viable if we use `Box` + `dyn` when using `-> impl Paginator`...

I wrote a more detailed description of this issue but there was a bug and my comment was removed. If GitHub support can't recover it I'll just write it again this weekend. Anyhow, this means that we are forced to use `Box` + `dyn` if we want this to work, which means a trait-based implementation isn't exactly optimal, meaning that perhaps we could explore other options.

marioortizmanero:
Unfortunately the diff is going to be a mess because I'm splitting `client.rs` into different files and `git` isn't able to keep track of that. So I would appreciate no breaking/huge changes to the library while I'm working on this, if possible.

marioortizmanero:
So I've got a couple examples working now with the new interface. You can take a look and let me know your thoughts/suggestions:

* https://github.com/ramsayleung/rspotify/blob/auth-rewrite/examples/album.rs
* https://github.com/ramsayleung/rspotify/blob/auth-rewrite/examples/current_user_recently_played.rs

marioortizmanero:
Something to consider as well is the naming, which is quite subjective so I'd like to hear opinions if possible:

* Should `ClientCredentialsSpotify` be renamed to `ClientCredsSpotify` to shorten the name? It's quite lengthy and I did shorten `CodeAuthorizationSpotify` to `CodeAuthSpotify`.
* Should `CodeAuthSpotify` be renamed to `AuthCodeSpotify`? It'd be more consistent with the real name, Authorization Code Flow

marioortizmanero:
This is more or less ready for a rough review, @ramsayleung. It's somewhat what I expected it to be now. I'd love some feedback before polishing it, as I wouldn't want to completely finish this PR if this architecture ends up not being any better.

In my honest opinion I think it's a great solution and I prefer it. But being more restrictive than before we might have more trouble in some cases. It currently has a few drawbacks, like having to `Box` pagination types as well, or not having official delegation yet, which might simplify the implementation a bit, but these will be fixed by Rust in the (possibly far) future.

What's your opinion?

src/code_auth.rs
```diff
@@ -0,0 +1,199 @@
...
+        data.insert(headers::GRANT_TYPE, headers::GRANT_AUTH_CODE);
+        data.insert(headers::REDIRECT_URI, oauth.redirect_uri.as_ref());
+        data.insert(headers::CODE, code);
+        data.insert(headers::SCOPE, scopes.as_ref());
+        data.insert(headers::STATE, oauth.state.as_ref());
+
+        let token = self.fetch_access_token(&data).await?;
+        self.token = Some(token);
+
+        self.write_token_cache()
```
ramsayleung, line 124: Same as above

src/code_auth.rs
```diff
@@ -0,0 +1,199 @@
...
+/// ![demo](https://raw.githubusercontent.com/ramsayleung/rspotify/master/doc/images/rspotify.gif)
+///
+/// Note: even if your script does not have an accessible URL, you will have to
+/// specify a redirect URI. It doesn't need to work, you can use
+/// `http://localhost:8888/callback` for example, which will also have the code
+/// appended like so: `http://localhost/?code=...`.
+///
+/// [reference]: https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-code-flow
+/// [example-main]: https://github.com/ramsayleung/rspotify/blob/master/examples/code_auth.rs
+/// [example-webapp]: https://github.com/ramsayleung/rspotify/tree/master/examples/webapp
```
ramsayleung, line 61: I think it will be better if we link reference with relative paths instead of HTTP links, which will be easily broken if I change username or repository

src/code_auth.rs
```diff
@@ -0,0 +1,199 @@
...
+pub struct CodeAuthSpotify {
+    pub creds: Credentials,
+    pub oauth: OAuth,
+    pub config: Config,
+    pub token: Option<Token>,
+    pub(in crate) http: HttpClient,
+}
+
+/// This client has access to the base methods.
+#[maybe_async(?Send)]
```
ramsayleung, line 73: Why would we need the `Send` trait?

src/endpoints/oauth.rs
```diff
@@ -0,0 +1,1202 @@
...
+        let (_, url) = params.find(|(key, _)| key == "code")?;
+        Some(url.to_string())
+    }
+
+    /// Tries to open the authorization URL in the user's browser, and returns
+    /// the obtained code.
+    ///
+    /// Note: this method requires the `cli` feature.
+    #[cfg(feature = "cli")]
+    fn get_code_from_user(&self, url: &str) -> ClientResult<String> {
```
ramsayleung, line 69: Not sure it's suitable to put `get_code_from_user` into `OAuthClient` trait, since we could get code without `token` and `cred` parameters. Same as `parse_response_code`.

src/endpoints/pagination/iter.rs
```diff
@@ -1,14 +1,15 @@
 //! Synchronous implementation of automatic pagination requests.
 
-use crate::client::{ClientError, ClientResult};
-use crate::model::Page;
+use crate::{model::Page, ClientError, ClientResult};
 
 /// Alias for `Iterator<Item = T>`, since sync mode is enabled.
-pub trait Paginator<T>: Iterator<Item = T> {}
-impl<T, I: Iterator<Item = T>> Paginator<T> for I {}
+pub type Paginator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
```
ramsayleung, line 6: I don't totally get it now why would we need to declare type `pub type Paginator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;` instead of the previous one ?
```rust
pub trait Paginator<T>: Iterator<Item = T> {}
impl<T, I: Iterator<Item = T>> Paginator<T> for I {}
```

Because we need some kinds of dynamic dispatch?

ramsayleung (approved):

ramsayleung:
Sorry for converting it from Draft to PR, my bad :(

What a huge PR, it takes me hours to give a rough review. The whole architecture of Auth process looks much clear and understandable than before, the current architecture looks good to me. Perhaps you could add some diagrams about the current Auth process.

> + Should ClientCredentialsSpotify be renamed to ClientCredsSpotify to shorten the name? It's quite lengthy and I did shorten CodeAuthorizationSpotify to CodeAuthSpotify.
> + Should CodeAuthSpotify be renamed to AuthCodeSpotify? It'd be more consistent with the real name, Authorization Code Flow
> + Should the endpoints module be renamed to clients? It's more fitting, as its contents aren't limited to endpoints anymore.

A big yes for all this suggestions.

By the ways, would you like splitting this PR into a Auth Rewrite PR and a Separate Crate PR, this PR has too many changes that it's a little hard to review line by line :)

Cargo.toml
```diff
@@ -25,19 +26,23 @@ resolver = "2"
 [dependencies]
 rspotify-macros = { path = "rspotify-macros", version = "0.10.0" }
 rspotify-model = { path = "rspotify-model", version = "0.10.0" }
+rspotify-http = { path = "rspotify-http", version = "0.10.0", default-features = false }
```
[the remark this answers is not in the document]
marioortizmanero: Because we need to manually specify which client to use. Perhaps `rspotify-http` shouldn't enable clients by default, and that should be done by `rspotify` itself?

src/client_creds.rs
```diff
@@ -0,0 +1,114 @@
...
+    /// Obtains the client access token for the app. The resulting token will be
+    /// saved internally.
+    #[maybe_async]
+    pub async fn request_token(&mut self) -> ClientResult<()> {
+        let mut data = Form::new();
+        data.insert(headers::GRANT_TYPE, headers::GRANT_CLIENT_CREDS);
+
+        self.token = Some(self.fetch_access_token(&data).await?);
+
+        self.write_token_cache()
```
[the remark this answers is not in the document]
marioortizmanero: I did implement cached tokens in this PR, but I've left refreshing ones for another.

rspotify-http/Cargo.toml
```diff
@@ -0,0 +1,51 @@
+[package]
+authors = ["Ramsay Leung <ramsayleung@gmail.com>"]
```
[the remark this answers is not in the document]
marioortizmanero: Sure!

src/client_creds.rs
```diff
@@ -0,0 +1,114 @@
...
+            // required scopes or it's expired.
+            None
+        } else {
+            Some(token)
+        }
+    }
+
+    /// Obtains the client access token for the app. The resulting token will be
+    /// saved internally.
+    #[maybe_async]
```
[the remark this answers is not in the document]
marioortizmanero: Yes. This function calls `write_token_cache` at the end so it will be written (only if cached tokens are enabled, as you may see in said function's definition)

src/code_auth.rs
marioortizmanero: I don't think we can link to the examples with relative paths, as they aren't uploaded to `docs.rs` (AFAIK). Or maybe I've misunderstood you?

src/code_auth.rs
marioortizmanero: This will probably explain it better than I can: https://docs.rs/async-trait/0.1.50/async_trait/#non-threadsafe-futures. I basically had to add `+ Send` to each generic type in the trait, so passing `(?Send)` to the macro does that automatically.

src/endpoints/pagination/iter.rs
marioortizmanero: A trait-based implementation of these pagination methods is only possible if we use `-> Box<dyn Paginator>` instead of  `-> impl Paginator`. The compiler will complain about it the same way we can't have `-> Future<...>`, for which we use `async_trait`. This is technically temporary and will be fixed in the future, but I'd give it *at least* a year until that happens, really. So we are basically forced to use `Box + dyn` if we want it to work for now. 

This is sub-optimal because there's a slight runtime overhead in comparison to generics, but on the other hand there should be less code bloat due to generics (not that either of these really matter, IMO).

marioortizmanero:
> What a huge PR, it takes me hours to give a rough review. The whole architecture of Auth process looks much clear and understandable than before, the current architecture looks good to me. Perhaps you could add some diagrams about the current Auth process.

Yes, it's a complete rewrite so the diff is just very hard to follow. I did want to add some diagrams but I wanted to make sure you were ok with it first of all. Here's a quick summary: we now have the implementations in traits: `rspotify::clients::BaseClient` and `rspotify::clients::OAuthClient`. We can then create custom structs which implement their own methods in order to follow whatever auth flow they're for, like `ClientCredsSpotify`.

This has some advantages and some inconvenients:

* GOOD: Initializing a struct for a specific flow is easier and more natural in comparison (no builder pattern needed, as you can see in the examples).
* GOOD: The implementation is somewhat cleaner, since we don't have to check if the struct has an `OAuth` struct in order to access certain methods. We *know* we can access the `OAuth` member in `OAuthClient` because of the `get_oauth` required implementation for the trait. This eliminates the need for an `InvalidAuth` error in the client.
* GOOD: The library is more type-safe, making it harder to make a mistake by calling an OAuth method in an unauthorized client.
* BAD: code reuse isn't that great. For example we can't override a method in `BaseClient` when implemeting its "child" `OAuthClient`, which would have been neat for `read_token_cache`, for example. This is fixable by using different names or by not implementing it in the trait and instead doing so in the client struct itself, but it's more of a workaround.

  I've also had a bit of trouble sharing behavior between OAuth clients, namely `AuthCodeSpotify` and `AuthCodePkceSpotify`. I didn't know if e.g. `prompt_for_token` should go in the trait or in the client itself, and this happened for quite a few methods. I've ended up adding the methods to the trait if possible in order to reuse more code, but perhaps they should go in a different trait, as they might not fit as great as I thought. I don't know, it's subject to change.

  This whole architecture would probably be nicer to implement if we had delegation in Rust. But that's not supported officially as of now (there are plans for it, though). We *could* use a delegation macro but I didn't find the problems I mentioned earlier important enough. Not sure what you think about that.

If you have any doubts do let me know. I know it's hard to follow the first time after a rewrite.

src/endpoints/oauth.rs
marioortizmanero: But `get_code_from_user` doesn't really make sense if you're not using oauth, right? You need the code for the authorization flows, which is not needed in a base client.

Cargo.toml
```diff
@@ -73,49 +66,44 @@ env-file = ["dotenv"]
...
+reqwest-native-tls-vendored = ["rspotify-http/reqwest-native-tls-vendored"]
 # Same for ureq.
-ureq-rustls-tls = ["ureq/tls"]
+ureq-rustls-tls = ["rspotify-http/ureq-rustls-tls"]
 
 # Internal features for checking async or sync compilation
-__async = ["async-trait", "async-stream", "futures"]
-__sync = ["maybe-async/is_sync"]
+__async = ["futures", "async-stream", "async-trait"]
+__sync = []
```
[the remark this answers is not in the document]
marioortizmanero: Yes, done.

ramsayleung:
> I didn't know if e.g. `prompt_for_token` should go in the trait or in the client itself, and this happened for quite a few methods. I've ended up adding the methods to the trait if possible in order to reuse more code, but perhaps they should go in a different trait, as they might not fit as great as I thought. I don't know, it's subject to change.

IMHO, moving methods like `prompt_for_token`, `parse_response_code`, `get_code_from_user` into a different trait would be a better idea, since they have weak relationship with Auth flow.

> How important is this? It would take me a bit of work to undo some changes and make two PRs instead.
It doesn't really matter, it's low priority.

marioortizmanero:
> IMHO, moving methods like prompt_for_token, parse_response_code, get_code_from_user into a different trait would be a better idea, since they have weak relationship with Auth flow.

How so? You only have to prompt for the user's authorization with the auth flows. Or am I missing something? I don't see why you'd use any of these in the client credentials flow.

ramsayleung:
> How so? You only have to prompt for the user's authorization with the auth flows. Or am I missing something? I don't see why you'd use any of these in the client credentials flow.

Oooh, I misunderstand something, you are right.

marioortizmanero:
Hello!

I've finally finished my finals, so I now have time to finish this. I've recreated the PR by splitting it up into four smaller PRs. The third one is still huge, but I don't think I can do anything about it; it separates the client into multiple ones, so the diff get messed up. As I mentioned there, the endpoints themselves aren't changed, so you can ignore them.

If this still isn't digestible enough please let me know, I don't want to kill you with my reviews, @ramsayleung haha.

Thanks a lot for the patience!

ramsayleung:
> If this still isn't digestible enough please let me know, I don't want to kill you with my reviews, @ramsayleung haha.

The good news is that there is a lot of reviews here, but it doesn't kill me. The bad news is it almost scares me out, haha :)

PR #209, opened 2021-05-02
Fix  not adding auth headers

luukvankooten:
## Description

Due to an issue with the method `current_user_playing_track` it does not add the auth headers required by spotify

## Motivation and Context

I have also created an issue #208 where the issue is explained.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

See #208 and run it with this patch it should return the `CurrentPlayingContext` struct.


marioortizmanero:
I realized this in #207 as well, so it's already fixed in the rewrite, in case it gets merged

PR #211, opened 2021-05-09
Use Github Action to automatically publish crate to crate.io

### LINKED ISSUE #210: [meta] could you tag crate releases
I'm always frustrated when I want to find the commit from which `cargo publish` was called and I find out there's no tag for it. ':-)

Can you please `git tag v0.x.x && git push origin v0.x.x` when you publish a version.

No need to make a GitHub release, just a tag would work.

**Additional context**
https://git-scm.com/book/en/v2/Git-Basics-Tagging

ramsayleung:
## Description

Adding a continuous deployment flow with Github Action that automatically publish everything to `crates.io` once a tag is published

## Motivation and Context

Fixed #210 

## Type of change

## How Has This Been Tested?

After this PR merged, creating a new tag named `0.10` and check if the Continuous Deployment flow works. Is there a better way to test?


marioortizmanero:
The thing is that this won't work after #207 because we'd need to publish multiple crates in that case.

marioortizmanero:
Note that #207 has already been merged, so you may continue your work on this now. No rush of course, as I see you're still busy with #224. I just wanted to update the status of this PR :)

As a tip, you might want to take a look at how already existing crates do this:

* Bevy uses a shell script: https://github.com/bevyengine/bevy/blob/main/tools/publish.sh
* Rust-Analyzer uses `cargo-workspaces`: https://github.com/rust-analyzer/rust-analyzer/blob/master/.github/workflows/publish.yml#L45

Bevy's solution is probably easier because I don't think we'll be modifying the set of crates in the repo anytime soon. If you have difficulties with the shell scripting let me know and I'll help you out.

marioortizmanero:
Yes they seem to release the new versions manually. In our case we just have to run the script in the CI flow.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,35 @@
...
+        shell: bash
+        run: |
+          # Not sure if it's necessary
+          git config --global user.email "runner@gha.local"
+          git config --global user.name "Github Action"
+
+          chmod +x "${GITHUB_WORKSPACE}/tool/publish.sh"
+
+          # Publishing crates
+          "${GITHUB_WORKSPACE}/tool/publish.sh"
```
marioortizmanero, line 35: this should be `tools`, not `tool` btw

marioortizmanero: also, can you make `publish.sh` executable? `chmod u+x publish.sh` should work fine

marioortizmanero:
No, sorry. What I meant is that you could make `publish.sh` an executable within the repo so that you don't need to run `chmod u+x publish.sh` in the CI flow

marioortizmanero:
Also I think the git config part isn't necessary, in other projects' CI I haven't added that and it worked fine. Afterwards we can merge.

marioortizmanero (approved):

PR #215, opened 2021-06-19
[Auth restructure 3] Split up `Spotify` into multiple clients

marioortizmanero:
Part 3 of splitting up #207, possibly the biggest of all changes. There's no way to reduce the number of changes here, really, as it requires moving most of the client's code into multiple files and different structures. But it doesn't really introduce that many changes. The endpoints themselves aren't changed at all and can be ignored, as they just were moved into traits.

The main library should be able to compile without problems after this PR.


src/auth_code_pkce.rs
```diff
@@ -0,0 +1,158 @@
...
+            .into_iter()
+            .collect::<Vec<_>>()
+            .join(" ");
+        payload.insert(headers::CLIENT_ID, &self.get_creds().id);
+        payload.insert(headers::RESPONSE_TYPE, headers::RESPONSE_CODE);
+        payload.insert(headers::REDIRECT_URI, &oauth.redirect_uri);
+        payload.insert(headers::SCOPE, &scope);
+        payload.insert(headers::STATE, &oauth.state);
+        // payload.insert(headers::CODE_CHALLENGE, todo!());
+        // payload.insert(headers::CODE_CHALLENGE_METHOD, "S256");
```
ramsayleung, line 153: According to the Spotify document, `code_challenge_method` and `code_challenge` are required when constructing the authorization URI. Why would you like to comment this code? It's still WIP?

ramsayleung (commented):
Except for the small nitpick I commented, everything else looks good to me. Great job!

src/auth_code_pkce.rs
marioortizmanero: I've left the PKCE implementation for a future PR

marioortizmanero:
One thing I'm still not sure about, though:

Maybe `prompt_for_token` should be implemented in the struct instead of in the trait? That way we could combine these two lines into one: https://github.com/ramsayleung/rspotify/blob/auth-rewrite-part4/examples/auth_code.rs#L44-L45.

ramsayleung:
>  That way we could combine these two lines into one: https://github.com/ramsayleung/rspotify/blob/auth-rewrite-part4/examples/auth_code.rs#L44-L45.

The link is broken for the part-4 PR has been merged.

marioortizmanero:
Here's the new link: https://github.com/ramsayleung/rspotify/blob/auth-rewrite-part3/examples/auth_code.rs#L44-L45

marioortizmanero:
Also you should wait to review all the PRs before merging any because otherwise the diff is going to have more unnecessary stuff (like what just happened after merging the fourth PR). If you didn't know you can always revert the `auth-rewrite-part4` and start from scratch.

ramsayleung:
> Also you should wait to review all the PRs before merging any because otherwise the diff is going to have more unnecessary stuff (like what just happened after merging the fourth PR). 

Yep, it makes sense. You could to merge master into this PR since it includes some changes have been merged into master by `auth-rewrite-part4`

marioortizmanero:
Sorry I don't quite understand your last comment

ramsayleung:
Sorry for the ambiguity, I mean, you could merge master into this `auth-rewrite-part3`, since `examples`, `tests`, `docs` and CI have been fixed, then it doesn't need to be shown on the diff. Is it clear for you now?

marioortizmanero:
You mean merge this into master? Well, it'd merge into auth-rewrite-part2 and we'd have the same problem in that case. Or have you reviewed it already?

ramsayleung:
> Or have you reviewed it already?

Not yet, I just took a rough review before, to check what has been modified. Then I will review this PR with caution, since it's a huge PR, it will take some time.

marioortizmanero:
> Alright, I could just manually ignore the diff which has been merged into master in PR auth-rewrite-part4

What I was saying before is that you can revert the auth-rewrite-part4 PR and apply it again after you're done:

image

But if you're ok with manually doing it then no problem.

src/lib.rs
```diff
@@ -51,144 +49,406 @@
...
+
+/// Structure that holds the required information for requests with OAuth.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OAuth {
+    pub redirect_uri: String,
+    /// The state is generated by default, as suggested by the OAuth2 spec:
+    /// [Cross-Site Request Forgery](https://tools.ietf.org/html/rfc6749#section-10.12)
+    pub state: String,
+    /// You could use macro [scopes!](crate::scopes) to build it at compile time easily
+    pub scope: HashSet<String>,
```
ramsayleung, line 408: Perhaps, we should rename it as `scopes`?

src/lib.rs
```diff
@@ -51,144 +49,406 @@
...
+            id: id.to_owned(),
+            secret: secret.to_owned(),
+        }
+    }
+
+    /// Parses the credentials from the environment variables
+    /// `RSPOTIFY_CLIENT_ID` and `RSPOTIFY_CLIENT_SECRET`. You can optionally
+    /// activate the `env-file` feature in order to read these variables from
+    /// a `.env` file.
+    pub fn from_env() -> Option<Self> {
```
ramsayleung, line 387: Perhaps we could return `ClientResult` for `from_env()` functions instead of `Option<Self>`?

src/lib.rs
```diff
@@ -51,144 +49,406 @@
...
+    /// requests, like [`artist_albums`](crate::clients::BaseClient). This
+    /// means that a request will be performed every `pagination_chunks` items.
+    /// By default this is [`DEFAULT_PAGINATION_CHUNKS`].
+    ///
+    /// Note that most endpoints set a maximum to the number of items per
+    /// request, which most times is 50.
+    pub pagination_chunks: u32,
+
+    /// TODO
+    pub token_cached: bool,
```
ramsayleung, line 231: I think `token_cached` is implemented to control whether to save the token cache, so it's fine to remove the `TODO` mark

src/clients/base.rs
```diff
@@ -0,0 +1,1054 @@
...
+    /// Parameters:
+    /// - user_id - the id of the usr
+    /// - limit  - the number of items to return
+    /// - offset - the index of the first item to return
+    ///
+    /// See [`Self::user_playlists_manual`] for a manually paginated version of
+    /// this.
+    ///
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-list-users-playlists)
+    fn user_playlists<'a>(
```
ramsayleung, line 1026: The `user_playlists` and `user_playlists_manual` endpoints should be put into the `oauth.rs` instead of base.rs. Since according to the document, 
private playlists are only retrievable for the current user and requires the `playlist-read-private` scope to have been authorized by the user

src/clients/base.rs
```diff
@@ -0,0 +1,1054 @@
...
+        convert_result(&result)
+    }
+
+    /// Gets basic profile information about a Spotify User.
+    ///
+    /// Parameters:
+    /// - user - the id of the usr
+    ///
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-users-profile)
+    async fn user(&self, user_id: &UserId) -> ClientResult<PublicUser> {
```
ramsayleung, line 444: This endpoint should be put into `oauth.rs` instead of `base.rs`, since it needs a valid access token obtained by authorization.

src/clients/base.rs
```diff
@@ -0,0 +1,1054 @@
...
+
+    /// Gets playlist of a user.
+    ///
+    /// Parameters:
+    /// - user_id - the id of the user
+    /// - playlist_id - the id of the playlist
+    /// - fields - which fields to return
+    ///
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-list-users-playlists)
+    async fn user_playlist(
```
ramsayleung, line 481: This endpoint should be put into `oauth.rs`

src/clients/base.rs
```diff
@@ -0,0 +1,1054 @@
...
+
+    /// Check to see if the given users are following the given playlist.
+    ///
+    /// Parameters:
+    /// - playlist_id - the id of the playlist
+    /// - user_ids - the ids of the users that you want to
+    /// check to see if they follow the playlist. Maximum: 5 ids.
+    ///
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-check-if-user-follows-playlist)
+    async fn playlist_check_follow(
```
ramsayleung, line 507: This endpoint should be put into `oauth.rs`

src/clients/pagination/iter.rs
```diff
@@ -1,14 +1,15 @@
...
+use crate::{model::Page, ClientError, ClientResult};
 
 /// Alias for `Iterator<Item = T>`, since sync mode is enabled.
-pub trait Paginator<T>: Iterator<Item = T> {}
-impl<T, I: Iterator<Item = T>> Paginator<T> for I {}
+pub type Paginator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
 
 /// This is used to handle paginated requests automatically.
-pub fn paginate<T, Request>(req: Request, page_size: u32) -> impl Iterator<Item = ClientResult<T>>
+pub fn paginate<'a, T: 'a, Request: 'a>(
```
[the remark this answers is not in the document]
marioortizmanero: When removing them:

```
error[E0310]: the parameter type `T` may not live long enough
  --> src/clients/pagination/iter.rs:23:5
   |
9  | pub fn paginate<T, Request>(
   |                 - help: consider adding an explicit lifetime bound...: `T: 'static`
...
23 |     Box::new(pages.flat_map(|result| ResultIter::new(result.map(|page| page.items.into_iter()))))
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ...so that the type `FlatMap<PageIterator<Request>, ResultIte
r<T, std::vec::IntoIter<T>>, [closure@src/clients/pagination/iter.rs:23:29: 23:96]>` will meet its required lifetime bounds
```

You can try declaring `T` and `Request` as `'static`, but you'll get this:

```
error[E0759]: `self` has lifetime `'a` but it needs to satisfy a `'static` lifetime requirement
   --> src/clients/base.rs:264:13
    |
258 |           &'a self,
    |           -------- this data with lifetime `'a`...
...
264 | /             move |limit, offset| {
265 | |                 self.artist_albums_manual(artist_id, album_type, market, Some(limit), Some(offset))
266 | |             },
    | |_____________^ ...is captured here...
    |
note: ...and is required to live as long as `'static` here
   --> src/clients/base.rs:263:9
    |
263 |         paginate(
    |         ^^^^^^^^
```

So `Request` borrows the client, which means it can't be `'static`. It makes sense; the `Paginator` can't live longer than the client itself because of this, so its lifetime has to be bounded. I could make `T` static, if anything. I've added a quick note to clarify the lifetimes in the `src/clients/pagination/mod.rs` file.

src/lib.rs
marioortizmanero: Yea I prefer that as well.

src/lib.rs
marioortizmanero: The thing is that `env::var` returns just an `Option`, so there's nothing else to report with a `Result` in that function.

I also wanted to do this for `Token::from_cache`, but then I would have to re-think this in `src/client_creds.rs`:

```rust
    /// Tries to read the cache file's token, which may not exist.
    ///
    /// Similarly to [`Self::write_token_cache`], this will already check if the
    /// cached token is enabled and return `None` in case it isn't.
    #[maybe_async]
    pub async fn read_token_cache(&self) -> Option<Token> {
        if !self.get_config().token_cached {
            return None;
        }

        let token = Token::from_cache(&self.get_config().cache_path)?;
        if token.is_expired() {
            // Invalid token, since it doesn't have at least the currently
            // required scopes or it's expired.
            // TODO: update with error?
            None
        } else {
            Some(token)
        }
    }
```

In that case we now have to return an error when the scopes don't match instead of (`Some`/`None`), and I'm not sure which one to pick. Perhaps we should add a new variant?

src/lib.rs
marioortizmanero: Done, thanks.

src/clients/base.rs
marioortizmanero: The thing with these endpoints you've mentioned is that, although it may work with OAuth, it *can also be used without it*. So they must be accessible from the base trait. Check out its reference:

image

And the footnote:

image

You can try it out here: https://developer.spotify.com/console/get-playlists/?user_id=wizzler&limit=&offset=

marioortizmanero:
Also, remember to only accept (and not merge) this PR after you're done. Then you can review the second PR, do the same, and after you agree with all of them merge them. That way the diff can be easy to follow.

src/lib.rs
ramsayleung: Well, I think `Some/None` is more reasonable, since we should return `None` when we could not find matched data, which is following the semantics of `Option` type.

src/clients/pagination/iter.rs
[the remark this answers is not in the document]
ramsayleung: Great, it makes sense.

src/clients/base.rs
ramsayleung: Sorry for not checking the footnote and its reference carefully, great point!

ramsayleung (approved):

ramsayleung:
Yeah, everything looks good to me, approve :)

PR #219, opened 2021-07-02
New CI

marioortizmanero:
This PR includes the new CI suggested in #213. It's merged over #215 to avoid merge conflicts later on.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,6 +54,7 @@ jobs:
             echo ">> Checking in $crate"
             cargo deadlinks --check-http --dir "$crate"
           done
+        continue-on-error: true
 
   test:
     name: Test and Lint for each Client
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,13 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v2
 
+      - name: Install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
       - name: Install cargo-deadlinks
         run: cargo install cargo-deadlinks
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,14 +27,14 @@ jobs:
           command: fmt
           args: --all -- --check
 
-  check-doc:
-    name: Check Dead Links in the Docs
+  docs:
+    name: Check Docs
     runs-on: ubuntu-latest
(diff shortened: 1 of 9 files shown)


marioortizmanero:
I've been taking a look at [Bevy's CI](https://github.com/bevyengine/bevy/blob/main/.github/workflows/ci.yml) and maybe we could copy some of their stuff? I'm taking a look right now.

marioortizmanero:
OK so I've added a flow that checks for broken links in the docs as well. This PR fixes all of them except for those that are impossible to fix until everything's merged. For example I added links to the new examples like https://github.com/ramsayleung/rspotify/blob/master/examples/auth_code.rs, but they won't work until the rewrite hits the `master` branch (they work with https://github.com/ramsayleung/rspotify/blob/auth-rewrite-part3/examples/auth_code.rs for now)

As to having a flow for each crate I'm not sure if it's worth it. Specially because the macros crate doesn't have that many tests, for example. I think having a flow for each client might be enough for our case.

marioortizmanero:
After a lot of debugging, turns out it's very complicated to have a clean run of `cargo-deadlinks` (i.e. without errors). For example, ours fails due to an external crate, https://github.com/tokio-rs/tracing/pull/1078.

Thus, I'm adding a `cargo-deadlinks` flow, but it will continue on error for now. This is what Bevy does as well, so it makes sense to me. It's still useful to check from time to time and it avoids unfixable errors.

This is ready for review, @ramsayleung

.github/workflows/ci.yml
```diff
@@ -27,51 +27,41 @@ jobs:
...
-      - name: Run cargo build
-        uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: -p rspotify -p rspotify-http -p rspotify-model -p rspotify-macros --no-default-features --features=${{ matrix.client }} --target ${{ matrix.target }} 
+      - name: Install cargo-deadlinks
+        run: cargo install cargo-deadlinks
+
+      - name: Build and Check
+        run: RUSTDOCFLAGS='-D warnings' cargo doc --features=cli,env-file
```
ramsayleung, line 48: Do we just check dead links for `cli` and `env-file` features, with ignoring other features?

ramsayleung:
I have to say, the `cargo-deadlinks` is great, it helps us find the broken links, instead of identifying it manually.

.github/workflows/ci.yml
marioortizmanero: The rest of the features don't add anything new to the docs.

ramsayleung (approved):

PR #222, opened 2021-07-07
Fix feature conflicts

### LINKED ISSUE #221: Avoiding conflicting features; allow multiple HTTP clients at once
Something that has been troubling me for a while is that we have conflicting features in rspotify, but [features in Rust must be *additive*](https://github.com/rust-lang/cargo/blob/master/src/doc/src/reference/features.md#feature-unification):

* Currently, the user can't enable multiple clients at the same time. If the features `client-ureq` and `client-reqwest` are enabled at the same time, they'll get hit by a compiler error warning them not to do so.
* It should work like `reqwest` does for its `blocking` feature. It's not a toggle. Instead, it enables the `blocking` module so that you may use it in combination of the async one.

This is important because the Rust compiler may merge features of a crate when it's duplicated in the dependency tree. If your crate has a dependency somewhere on rspotify with reqwest and at the same time somewhere else rspotify with ureq, that will never possibly work because you can't merge `client-ureq` with `client-reqwest`. This is explained a few times in [this post](https://www.reddit.com/r/rust/comments/ofm4nl/why_you_shouldnt_obsess_about_rust_features/), for example [here](https://www.reddit.com/r/rust/comments/ofm4nl/why_you_shouldnt_obsess_about_rust_features/h4dozda/).

I thought this was more complicated than it actually is (still not easy). We actually have a base HTTP trait already, so what if we make the Spotify client generic over it? Here's a playground link that explains it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=de0e9c0a596a5004b3e4af41071415d6. It might simplify a lot of things, really! The only downside is that the syntax to initialize the client will be a bit ugiler to specify the HTTP client. 

Related issues:
* https://github.com/rust-lang/cargo/issues/4328
* https://github.com/fMeow/maybe-async-rs/issues/6
* https://github.com/fMeow/arangors/issues/37

marioortizmanero:
Closes #221

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,11 +59,6 @@ jobs:
   test:
     name: Test and Lint for each Client
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        features:
-          - rspotify/cli,rspotify/env-file,rspotify/client-ureq,rspotify/ureq-rustls-tls,rspotify-http/client-ureq,rspotify-http/ureq-rustls-tls
-          - rspotify/cli,rspotify/env-file,rspotify/client-reqwest,rspotify/reqwest-rustls-tls,rspotify-http/client-reqwest,rspotify-http/reqwest-rustls-tls
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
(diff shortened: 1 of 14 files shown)


marioortizmanero:
The concept of being generic over the http client can definitely work. The wall I've just hit though is that `maybe_async` itself is a toggle; you can't have both async and blocking at the same time. So I have to find a way to make that possible. Also, the paginators conflict.

marioortizmanero:
Superseded by #253 with a different idea

PR #224, opened 2021-07-10
Support re-authenticate automatically and refresh token when it expired.

### LINKED ISSUE #223: Implement cache token and refresh token
Since we use explicitly individual functions to control whether to save the token into Json file:

+ `refresh_user_token_without_cache`
+ `refresh_user_token`(with cache implicitly)

Honestly speaking, it's clumsy but work. We are planning to improve it, since we have discussed that [making it as a feature](https://github.com/ramsayleung/rspotify/issues/135) is not a good idea, I prefer to use configuration field to control. Then @marioortizmanero already added two fields to control this behavior:

```rust
pub struct Config {
    pub token_cached: bool,

    /// TODO
    pub token_refreshing: bool,
}
```

Since the skeleton has been constructed, I would like to supply the flesh to make the body complete.

ramsayleung:
## Description

Fixed #223 and #4, since @marioortizmanero already implemented the cache token feature, what I need to do is implement the re-authenticate feature and refresh token when it expired.

To be honest, it's a little hard to implement than I expect before. The major problem is that to maintain a mutable token inside the immutable client.

+ `Cell<T>` doesn't work as expected, since it requires the types that implement `Copy` trait, but `Token` doesn't implement the `Copy` trait and has no way to do so. Since the String type doesn't implement.
+ `Mutex<T>` seems to be unnecessary, for that we don't need to handle the shared-state between threads(or we should do so?)
+ `RefCell<T>` is great, it looks pretty suitable for our case, but it also follows Rust borrow-check rules, even though the check delays to runtime:
> At any given time, you can have either (but not both of) one mutable reference or any number of immutable references.

So the following code will fail:

```rust
let c = RefCell::new(5);
let m = c.borrow();

let b = c.borrow_mut(); // this causes a panic
```
Thus, we need to take good care of the borrowed immutable reference and borrowed mutable reference.

And there are a little `TODO` items I left to discuss, I am not sure which way should we choose:

+ I change the function signature of `request_token/refresh_token` from `&mut self` to `&self`, so we have an immutable client now. But the trade-off is we use `borrow_mut` internally, which means we leave the immutable or mutable borrows checked at runtime, we are in charge of compiler's jobs. It looks like a landmine, will bomb someday by mistake.

## Motivation and Context

+ #176
+ #4 
+ #223  

## Dependencies 

Nope

## Type of change

## How Has This Been Tested?

- [ ] Test A: examples/with_refresh_token

### DIFF
--- a/src/client_creds.rs
+++ b/src/client_creds.rs
@@ -39,7 +39,7 @@ impl BaseClient for ClientCredsSpotify {
         self.token.borrow()
     }
 
-    async fn get_token_mut(&mut self) -> RefMut<Option<Token>> {
+    async fn get_token_mut(&self) -> RefMut<Option<Token>> {
         self.token.borrow_mut()
     }
 
--- a/src/client_creds.rs
+++ b/src/client_creds.rs
@@ -6,6 +6,7 @@ use crate::{
 };
 
 use maybe_async::maybe_async;
+use std::cell::{Ref, RefCell, RefMut};
 
 /// The [Client Credentials Flow][reference] client for the Spotify API.
 ///
@@ -23,7 +24,7 @@ use maybe_async::maybe_async;
 pub struct ClientCredsSpotify {
     pub config: Config,
     pub creds: Credentials,
-    pub token: Option<Token>,
+    pub token: RefCell<Option<Token>>,
     pub(in crate) http: HttpClient,
 }
 
@@ -34,12 +35,12 @@ impl BaseClient for ClientCredsSpotify {
         &self.http
     }
 
-    async fn get_token(&self) -> Option<&Token> {
-        self.token.as_ref()
+    async fn get_token(&self) -> Ref<Option<Token>> {
+        self.token.borrow()
     }
 
-    async fn get_token_mut(&mut self) -> Option<&mut Token> {
-        self.token.as_mut()
+    async fn get_token_mut(&mut self) -> RefMut<Option<Token>> {
+        self.token.borrow_mut()
     }
 
     fn get_creds(&self) -> &Credentials {
@@ -66,7 +67,7 @@ impl ClientCredsSpotify {
     /// as the client credentials aren't known.
     pub fn from_token(token: Token) -> Self {
         ClientCredsSpotify {
-            token: Some(token),
+            token: RefCell::new(Some(token)),
             ..Default::default()
         }
     }
@@ -108,7 +109,7 @@ impl ClientCredsSpotify {
         let mut data = Form::new();
         data.insert(headers::GRANT_TYPE, headers::GRANT_CLIENT_CREDS);
 
-        self.token = Some(self.fetch_access_token(&data).await?);
+        *self.token.borrow_mut() = Some(self.fetch_access_token(&data).await?);
 
         self.write_token_cache().await
     }
--- a/src/client_creds.rs
+++ b/src/client_creds.rs
@@ -28,16 +28,17 @@ pub struct ClientCredsSpotify {
 }
 
 /// This client has access to the base methods.
+#[maybe_async(?Send)]
 impl BaseClient for ClientCredsSpotify {
     fn get_http(&self) -> &HttpClient {
         &self.http
     }
 
-    fn get_token(&self) -> Option<&Token> {
+    async fn get_token(&self) -> Option<&Token> {
         self.token.as_ref()
     }
 
-    fn get_token_mut(&mut self) -> Option<&mut Token> {
+    async fn get_token_mut(&mut self) -> Option<&mut Token> {
         self.token.as_mut()
     }
 
@@ -109,6 +110,6 @@ impl ClientCredsSpotify {
 
         self.token = Some(self.fetch_access_token(&data).await?);
 
-        self.write_token_cache()
+        self.write_token_cache().await
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -363,6 +363,11 @@ impl Token {
         self.expires_at
             .map_or(true, |x| Utc::now().timestamp() > x.timestamp())
     }
+
+    /// Check if the token is ready to re-authenticate automatically
+    pub fn is_capable_to_reauth(&self) -> bool {
+        self.is_expired() && self.refresh_token.is_some()
+    }
 }
 
 /// Simple client credentials object for Spotify.
--- a/examples/webapp/Cargo.toml
+++ b/examples/webapp/Cargo.toml
@@ -5,8 +5,9 @@ authors = ["Ramsay <ramsayleung@gmail.com>"]
 edition = "2018"
 
 [dependencies]
-rocket = "0.4.5"
-rocket_contrib = { version = "0.4.5", features = ["tera_templates"] }
+
+rocket = { version = "0.5.0-rc.1", features = ["json"] }
+rocket_dyn_templates={ version = "0.1.0-rc.1", features = ["tera", "handlebars"] }
 getrandom = "0.2.0"
 # Rocket is synchronous, so this uses the `ureq` client
 rspotify = { path = "../..", features = ["client-ureq", "ureq-rustls-tls"], default-features = false }
--- a/examples/webapp/src/main.rs
+++ b/examples/webapp/src/main.rs
@@ -2,26 +2,22 @@
 //! real-world web app, you should store it in a database instead. In that case
 //! you can disable `token_cached` in the `Config` struct passed to the client
 //! when initializing it to avoid using cache files.
-
-#![feature(proc_macro_hygiene, decl_macro)]
-
 #[macro_use]
 extern crate rocket;
-
 use getrandom::getrandom;
-use rocket::http::{Cookie, Cookies};
-use rocket::response::Redirect;
-use rocket_contrib::json;
-use rocket_contrib::json::JsonValue;
-use rocket_contrib::templates::Template;
-use rspotify::{scopes, AuthCodeSpotify, OAuth, Credentials, Config, prelude::*, Token};
+use rocket::http::{Cookie, CookieJar, SameSite};
+use rocket::serde::json::json;
+use rocket::{launch, Build};
+use rocket::{routes, Rocket};
+use rocket_dyn_templates::handlebars::JsonValue;
 
+use rocket::response::Redirect;
+use rocket_dyn_templates::Template;
+use rspotify::{prelude::*, scopes, AuthCodeSpotify, Config, Credentials, OAuth, Token};
+
+use std::cell::RefCell;
 use std::fs;
-use std::{
-    collections::HashMap,
-    env,
-    path::PathBuf,
-};
+use std::{collections::HashMap, env, path::PathBuf};
 
 #[derive(Debug, Responder)]
 pub enum AppResponse {
@@ -45,16 +41,16 @@ fn generate_random_uuid(length: usize) -> String {
         .collect()
 }
 
-fn get_cache_path(cookies: &Cookies) -> PathBuf {
+fn get_cache_path(cookies: &CookieJar<'_>) -> PathBuf {
     let project_dir_path = env::current_dir().unwrap();
     let mut cache_path = project_dir_path;
     cache_path.push(CACHE_PATH);
-    cache_path.push(cookies.get("uuid").unwrap().value());
+    cache_path.push(cookies.get_pending("uuid").unwrap().value());
 
     cache_path
 }
 
-fn create_cache_path_if_absent(cookies: &Cookies) -> PathBuf {
+fn create_cache_path_if_absent(cookies: &CookieJar<'_>) -> PathBuf {
     let cache_path = get_cache_path(cookies);
     if !cache_path.exists() {
         let mut path = cache_path.clone();
@@ -64,7 +60,7 @@ fn create_cache_path_if_absent(cookies: &Cookies) -> PathBuf {
     cache_path
 }
 
-fn remove_cache_path(mut cookies: Cookies) {
+fn remove_cache_path(cookies: &CookieJar<'_>) {
     let cache_path = get_cache_path(&cookies);
     if cache_path.exists() {
         fs::remove_file(cache_path).unwrap()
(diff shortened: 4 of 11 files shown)


marioortizmanero (commented):
The thing is, I also thought of using a `RefCell`, but the problem is that with it we disallow concurrency for the client. I made a quick example which can be ran with:

```
cargo run --example threading --features=cli,env-file,client-ureq,ureq-rustls-tls --no-default-features
```

Saved as `examples/threading.rs`:

```rust
use rspotify::{model::Id, prelude::*, ClientCredsSpotify, Credentials};
use std::thread;

fn main() {
    let creds = Credentials::from_env().unwrap();

    let mut spotify = ClientCredsSpotify::new(creds);

    spotify.request_token().unwrap();

    let birdy_uri = Id::from_uri("spotify:album:0sNOF9WDwhWunNAHPD3Baj").unwrap();
    let h1 = thread::spawn(|| {
        let albums = spotify.album(birdy_uri);
        println!("Response 1: {:#?}", albums);
    });

    let h2 = thread::spawn(|| {
        let albums = spotify.album(birdy_uri);
        println!("Response 2: {:#?}", albums);
    });

    h1.join();
    h2.join();
}
```

Which will spit out the following error:

```
error[E0277]: `RefCell<Option<Token>>` cannot be shared between threads safely
   --> examples/threading.rs:12:14
    |
12  |     let h1 = thread::spawn(|| {
    |              ^^^^^^^^^^^^^ `RefCell<Option<Token>>` cannot be shared between threads safely
    |
   ::: /home/mario/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/thread/mod.rs:619:8
    |
619 |     F: Send + 'static,
    |        ---- required by this bound in `spawn`
    |
    = help: within `ClientCredsSpotify`, the trait `Sync` is not implemented for `RefCell<Option<Token>>`
    = note: required because it appears within the type `ClientCredsSpotify`
    = note: required because of the requirements on the impl of `Send` for `&ClientCredsSpotify`
    = note: required because it appears within the type `[closure@examples/threading.rs:12:28: 15:6]`

For more information about this error, try `rustc --explain E0277`.

```

I think we should allow concurrency in the client, so perhaps a `Mutex` or `RwLock` would be best for this case. I would expect the latter to be a better option because we don't need to lock to read the token.

Similarly, it weirds me out that `request_token` and similars don't take `&mut self` anymore... But I don't think we can do anything about that.

ramsayleung:
> I think we should allow concurrency in the client, so perhaps a Mutex or RwLock would be best for this case. I would expect the latter to be a better option because we don't need to lock to read the token.

Great point!

src/client_creds.rs
```diff
@@ -23,22 +24,23 @@ use maybe_async::maybe_async;
...
+#[maybe_async(?Send)]
 impl BaseClient for ClientCredsSpotify {
     fn get_http(&self) -> &HttpClient {
         &self.http
     }
 
-    fn get_token(&self) -> Option<&Token> {
-        self.token.as_ref()
+    async fn get_token(&self) -> Ref<Option<Token>> {
+        self.token.borrow()
```
marioortizmanero, line 39: By the way why does the client_credentials auth flow not re-authenticate?

marioortizmanero (commented):
I was wondering if we could include `refresh_token` in the base trait (only the requirement, the implementation would still go in the clients) so that `auto_reauth` can be implemented in the trait instead?

src/lib.rs
```diff
@@ -363,6 +368,11 @@ impl Token {
         self.expires_at
             .map_or(true, |x| Utc::now().timestamp() > x.timestamp())
     }
+
+    /// Check if the token is ready to re-authenticate automatically
+    pub fn is_capable_to_reauth(&self) -> bool {
```
marioortizmanero, line 373: Perhaps `can_reauth` would be a better name?

src/lib.rs
```diff
@@ -228,9 +228,14 @@ pub struct Config {
     /// request, which most times is 50.
     pub pagination_chunks: u32,
 
+    /// Whether or not to save the authentication token into a JSON file,
+    /// then reread token from JSON file when launch the program without
```
marioortizmanero, line 232: typos: 

* "then reread token" -> "then reread the token"
* "when launch the program" -> "when launching the program"

src/lib.rs
ramsayleung: Yes, it's a better name. 

It reminds me that since `get_token` and `get_token_mut` functions not only simply returning the `Token`, but also re-authenticate automatically. The semantics of these functions has changed, it's better to have a more proper name to show the implicit flow. There some propositions:

+ `check_and_get_token`
+ `reauth_and_get_token`

Any other suggestion?

src/client_creds.rs
ramsayleung: According to the [official document](https://developer.spotify.com/documentation/general/guides/authorization-guide/#authorization-flows), the client_credentials auth flow doesn't support access token refresh. If we want to implement re-authenticate for client credential auth flow, we should figure out another way to go.

Sorry for not mentioning this point.

marioortizmanero: Since this doesn't require user interaction, could it perhaps be refreshed by just authenticating the usual way?

src/lib.rs
marioortizmanero: Hmmm I get your point, but I think `get_token` isn't that bad of an idea. The token refreshing may not even be activated, so `get_token` still makes sense. Is there any occasion where you may want to get the token without refreshing it?

marioortizmanero:
I suggest you to apply changes only after #232 is merged here to avoid conflicts

marioortizmanero:
Should we make the field `token` in the client private so that it's not accessed directly without `get_token`? Not sure myself.

marioortizmanero:
And something else I'm not sure about: should the token be `RwLock<Option<Token>>`, or `Option<RwLock<Token>>`?

ramsayleung:
> I suggest you to apply changes only after #232 is merged here to avoid conflicts

Yep, And I think #232 is ready to merge :)

marioortizmanero:
Yes, you can merge it if you like.

src/lib.rs
ramsayleung: Your point is reasonable, we could keep the name. But I think it's a little weird that the `get_token` is a `maybe_async` function and `get_token_mut` is `sync` function, it breaks the consistency.

dylanhand:
Hi and thanks for all the great work on this library!

Any movement on this issue? Looks like it's just awaiting review from @marioortizmanero?

Looking forward to getting this re-authentication fix in my project :)

marioortizmanero:
Hi! I arrived back from vacation yesterday, so I can keep working on rspotify now.

As far as I remember, what was pending was, in order of its priority:

1. Add the automatic authentication to the client credentials flow as well (whenever @ramsayleung has time)
2. Figure out how to avoid code duplication for `auto_reauth`
3. Decide the new names for `get_token[_mut]`

dylanhand:
Great :) Thanks again for all of your and @ramsayleung's hard work on this!

ramsayleung:
> Hi! I arrived back from vacation yesterday, so I can keep working on rspotify now.
> 
> As far as I remember, what was pending was, in order of its priority:
> 
>     1. Add the automatic authentication to the client credentials flow as well (whenever @ramsayleung has time)
> 
I have added the automatic authentication to the client credentials flow in this commit: [`d0826a6` (#224)](https://github.com/ramsayleung/rspotify/pull/224/commits/d0826a69bec0d39342a6a64f548bf4496888567a), you could take a look at when you have time

>     2. Figure out how to avoid code duplication for `auto_reauth` (we should move as much code as possible to the trait)
> 
>     3. Decide the new names for `get_token[_mut]`

As for these two points, we could discuss how to figure out a better solution :)

marioortizmanero:
Oh wow. Turns out that it's impossible to merge master with this after #237... `RwLock` is not `Send`, so the client isn't either. We have two options:

* Make the client non-`Send`
* Use a `Mutex` instead of a `RwLock`

So perhaps `Mutex` for the token is the way to go in the end?

We should apply these changes after merging #239 to avoid more conflicts.

marioortizmanero:
@kebabtent, as the opener of #236, could you provide some insight as to why it's important that the client is `Send`? Perhaps your use case will help us decide.

ramsayleung:
We could keep working on this PR, even though no more response from @kebabtent

> We have two options:

> + Make the client non-Send
> + Use a `Mutex` instead of a `RwLock`

I personally prefer the second option, which allows concurrency for the client.

marioortizmanero:
Agreed. Can you change it so that this uses a mutex then?

ramsayleung:
> Yes, I will implement it with Mutex on this weekend

Well, it's a little harder than I expect to implement, the combination of async code and concurrency code is complex
```rust
error[E0053]: method `get_token` has an incompatible type for trait
  --> src/auth_code_pkce.rs:46:5
   |
46 |     async fn get_token(&self) -> MutexGuard<Option<Token>> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + std::marker::Send`, found trait `futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>>`
   |
  ::: src/clients/base.rs:49:5
   |
49 |     async fn get_token(&self) -> MutexGuard<Option<Token>>;
   |     ------------------------------------------------------- type in trait
   |
   = note: expected fn pointer `fn(&'life0 AuthCodePkceSpotify) -> Pin<Box<(dyn futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + std::marker::Send + 'async_trait)>>`
              found fn pointer `fn(&'life0 AuthCodePkceSpotify) -> Pin<Box<(dyn futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + 'async_trait)>>`

error[E0053]: method `refetch_token` has an incompatible type for trait
  --> src/auth_code_pkce.rs:69:5
   |
69 |     async fn refetch_token(&self) -> ClientResult<Option<Token>> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `futures::Future<Output = Result<std::option::Option<Token>, ClientError>> + std::marker::Send`, found trait `futures::Future<Output = Result<std::option::Option<Token>, ClientError>>`
   |
  ::: src/clients/base.rs:53:5
   |
53 |     async fn refetch_token(&self) -> ClientResult<Option<Token>>;
   |     ------------------------------------------------------------- type in trait
   |
   = note: expected fn pointer `fn(&'life0 AuthCodePkceSpotify) -> Pin<Box<(dyn futures::Future<Output = Result<std::option::Option<Token>, ClientError>> + std::marker::Send + 'async_trait)>>`
              found fn pointer `fn(&'life0 AuthCodePkceSpotify) -> Pin<Box<(dyn futures::Future<Output = Result<std::option::Option<Token>, ClientError>> + 'async_trait)>>`

error[E0053]: method `get_token` has an incompatible type for trait
  --> src/client_creds.rs:39:5
   |
39 |     async fn get_token(&self) -> MutexGuard<Option<Token>> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + std::marker::Send`, found trait `futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>>`
   |
  ::: src/clients/base.rs:49:5
   |
49 |     async fn get_token(&self) -> MutexGuard<Option<Token>>;
   |     ------------------------------------------------------- type in trait
   |
   = note: expected fn pointer `fn(&'life0 ClientCredsSpotify) -> Pin<Box<(dyn futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + std::marker::Send + 'async_trait)>>`
              found fn pointer `fn(&'life0 ClientCredsSpotify) -> Pin<Box<(dyn futures::Future<Output = std::sync::MutexGuard<'life0, std::option::Option<Token>>> + 'async_trait)>>`

error[E0053]: method `refetch_token` has an incompatible type for trait
  --> src/client_creds.rs:65:5
   |
65 |     async fn refetch_token(&self) -> ClientResult<Option<Token>> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected trait `futures::Future<Output = Result<std::option::Option<Token>, ClientError>> + std::marker::Send`, found trait `futures::Future<Output = Result<std::option::Option<Token>, ClientError>>`
   |

```

I am still working on it.

marioortizmanero:
Yeah, we should use `Arc<Mutex<T>>`. Also, if we have a `Mutex` we only need a single getter; `get_token_mut` is redundant. Perhaps we could call the method `lock_token` instead to point out that it returns a locked mutex with interior mutability.

ramsayleung:
> Yeah, we should use Arc<Mutex<T>>. Also, if we have a Mutex we only need a single getter; get_token_mut is redundant. Perhaps we could call the method lock_token instead to point out that it returns a locked mutex with interior mutability.

Yeah, I totally agree with you, it seems the only way to go, to return a locked `Mutex`. The caller tries to acquire the lock when accessing or updating date.

But the problem is the function will become less ergonomic to use, since we expose the complexity to the function caller(even though the `get_token` will only be used inside the `rspotify`)

marioortizmanero:
I agree that it's ugly as hell, but as you said it's literally the only way. Now it should be as easy as fixing the compilation errors and cleaning up the docs (mainly the ones in `BaseClient::get_token`

ramsayleung:
I am aware of that it's still much complicated with `Arc<Mutex<T>>` than I supposed, because we can't lock across `.await`, since 
the switching among threads can happen at awaiting points. For more detail, you could check this [Reddit thread](https://www.reddit.com/r/rust/comments/guivuf/mutex_in_async_world/).

Well, the synchronous `Mutex`can't be used across `.await`, it's reasonable to use asynchronous `Mutex`, then a new problem occurs: the function signature of these asynchronous `Mutex`es are different with the synchronous:
- asynchronous version:
  + [`futures::sync::Mutex`](https://docs.rs/futures/0.3.4/futures/lock/struct.Mutex.html): `pub fn lock(&self) -> MutexLockFuture<T>` 
  + [`async_mutex::Mutex`](https://docs.rs/async-mutex/1.4.0/async_mutex/struct.Mutex.html): `pub async fn lock<'_, '_>(&'_ self) -> MutexGuard<'_, T>`
- ssynchronous version:
  + [`std::sync::Mutex`](https://doc.rust-lang.org/std/sync/struct.Mutex.html): `pub fn lock(&self) -> LockResult<MutexGuard<'_, T>>`

It's not compatible to switch between `async` and `sync` with `maybe_async` crate.

marioortizmanero:
Damn it does seem more complicated than I thought. I'll take a look and I'll let you know but it'll probably take me some time to fully understand everything :P

ramsayleung:
Closed over #262

PR #232, opened 2021-07-15
Refresh token with Rwlock

marioortizmanero:
This is my attempt at #224 with a RwLock to enable concurrency.

The main drawback is that we can't derive `Clone` for the clients. But it allows concurrency for the client, which may be a must for blocking clients?

### DIFF
--- a/src/auth_code.rs
+++ b/src/auth_code.rs
@@ -91,14 +91,14 @@ impl BaseClient for AuthCodeSpotify {
         self.auto_reauth()
             .await
             .expect("Failed to re-authenticate automatically, please authenticate");
-        self.token.read().unwrap()
+        self.token.read().expect("Failed to read token; the lock has been poisoned")
     }
 
     async fn get_token_mut(&self) -> RwLockWriteGuard<Option<Token>> {
         self.auto_reauth()
             .await
             .expect("Failed to re-authenticate automatically, please authenticate");
-        self.token.write().unwrap()
+        self.token.write().expect("Failed to write token; the lock has been poisoned")
     }
 }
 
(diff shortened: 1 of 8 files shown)


marioortizmanero:
Also, I've tried to allow shared-state concurrency with `RwLock` instead of `Mutex` just because I assume most times the token is going to be read and not written, in which case its access is truly concurrent (without blocking) most times.

But if someone thinks `Mutex` would actually be more performant please do let me know.

ramsayleung:
Well, it's a great idea to use `RWLock` to support shared-state between threads, I like this idea :)

src/auth_code.rs
```diff
@@ -85,18 +87,18 @@ impl BaseClient for AuthCodeSpotify {
         &self.config
     }
 
-    async fn get_token(&self) -> Ref<Option<Token>> {
+    async fn get_token(&self) -> RwLockReadGuard<Option<Token>> {
         self.auto_reauth()
             .await
             .expect("Failed to re-authenticate automatically, please authenticate");
-        self.token.borrow()
+        self.token.read().unwrap()
```
ramsayleung, line 94: Perhaps we could handle the exception more elegantly than `unwrap`?

marioortizmanero: sure!

marioortizmanero:
I was thinking that `get_token_mut` doesn't really need to refresh the token automatically, right? It's going to be overwritten anyway. I'll modify my code so that it doesn't refresh the token.

ramsayleung:
> I was thinking that get_token_mut doesn't really need to refresh the token automatically, right? It's going to be overwritten anyway. I'll modify my code so that it doesn't refresh the token.

Yes, you make the point. I will remove the refresh flow from the `get_token_mut` function

marioortizmanero:
Don't worry I already did that.

PR #238, opened 2021-08-07
Attempt 1 at simplifying the token reauth

marioortizmanero:
This is a WIP attempt to simplify the code at #224


marioortizmanero:
Ok so I've tried to reduce the boilerplate code in #224 and I've come up with this.

I've basically converted `auto_reauth` into a generic function that can be used by **all** the clients. It only needs to be passed a function to refresh the token, and `auto_reauth` will automatically check if the token should be refreshed at all.

* https://github.com/ramsayleung/rspotify/pull/238/commits/e081904a00154cb8e051b0b49b154ec0efb6a54e is independent from this PR and should be applied to #224 regardless of whether this is merged or not; it simplifies the `auto_reauth` function for the OAuth clients and makes it nicer to read IMO.
* https://github.com/ramsayleung/rspotify/pull/238/commits/bc1728ceade1eef6108b1e6495aa231a88dd985c contains all the changes I've explained before.

After implementing this I think it's probably overengineering; it might be best to just copy-paste some parts of the code. Thoughts?

PR #239, opened 2021-08-07
Attempt 2 at simplifying the token reauth

marioortizmanero:
This is a WIP attempt to simplify the code at #224

### DIFF
--- a/examples/with_refresh_token.rs
+++ b/examples/with_refresh_token.rs
@@ -68,15 +68,6 @@ async fn main() {
         .prompt_for_token(&url)
         .await
         .expect("couldn't authenticate successfully");
-    let refresh_token = spotify
-        .get_token()
-        .await
-        .as_ref()
-        .unwrap()
-        .refresh_token
-        .as_ref()
-        .unwrap()
-        .clone();
     do_things(spotify).await;
 
     // At a different time, the refresh token can be used to refresh an access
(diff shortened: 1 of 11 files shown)


marioortizmanero:
Ok I like this approach more than #238. It just moves `refetch_token` and `refresh_token` to the base trait, `auto_reauth` to the oauth one, and simplifies `auto_reauth`. So nothing special.

Let me know what you think, @ramsayleung

ramsayleung (approved):
Look good to me.

ramsayleung:
Perhaps you could add a diagram of trait hierarchy, I get loss from time to time :)

examples/with_refresh_token.rs
```diff
@@ -95,7 +86,7 @@ async fn main() {
     println!(">>> Session three, running some requests:");
     let spotify = AuthCodeSpotify::new(creds.clone(), oauth.clone());
     spotify
-        .refresh_token(&refresh_token)
```
ramsayleung, line 98: Oooh, it turns out that it's impossible to `refresh_token` without the `token.refresh_token`.

When we try to `refresh_token` by `refetch_token` function without `token.refresh_token`, it will simply `panicked at 'Rspotify not authenticated', rspotify/src/clients/base.rs:104:55`, because the `auth_headers` function need an authenticated token.

PR #240, opened 2021-08-08
Add trait hierarchy diagram.

### LINKED ISSUE #173: Restructure the authentication process

ramsayleung:
## Description

Adding a diagram with the full hierarchy in the docs, making it easier for newcomers to contribute

## Motivation and Context

Fixed #173 https://github.com/ramsayleung/rspotify/issues/173#issuecomment-785297882

## Type of change


marioortizmanero:
I love it! Thanks for taking the time to create the diagram.

marioortizmanero (approved):

PR #241, opened 2021-08-20
Fix IDs

### LINKED ISSUE #203: Endpoints with lists of Ids don't accept multiple of them
**Describe the bug**

```rust

#[tokio::main]
async fn main() {
    env_logger::init();

        .credentials(creds)
        .build()
        .unwrap();

    let uris = &[
    ];

}
```

You get:

```
  --> examples/broken.rs:20:9
   |
Track`
   |

```

marioortizmanero:
## Description

This closes #203. It also includes a few more changes about IDs, like using them in `rspotify-model`, more tests and documenting them thoroughly.

## Motivation and Context

I've fixed *some* of the issues I found about IDs. I still need to figure out :

* how to take multiple types of IDs 
* how to fix some compiler issue about `Deserialize<'_>`

### DIFF
--- a/rspotify-model/src/idtypes.rs
+++ b/rspotify-model/src/idtypes.rs
@@ -23,6 +23,47 @@
 //!    still needs to be a variant from [`crate::Type`])
 //! 2. They may be used anywhere, as they implement every single trait in this
 //!    module.
+//!
+//! ## Examples
+//!
+//! You may let the type inferrer guess what kind of ID you're dealing with by
+//! using [`Id`]:
+//!
+//!
+//! ```
+//! fn pause_track(id: &TrackId) { /* ... */ }
+//!
+//! let id = Id::from_id("4iV5W9uYEdYUVa79Axb7Rh").unwrap();
+//! pause_track(id); // this function takes a `TrackId`, so `id` must be one
+//! ```
+//!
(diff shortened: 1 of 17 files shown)


marioortizmanero:
Closing this in favor of #243

rspotify-model/src/idtypes.rs
```diff
@@ -17,51 +88,56 @@ mod private {
...
-pub type ArtistIdBuf = IdBuf<Artist>;
-pub type AlbumIdBuf = IdBuf<Album>;
-pub type TrackIdBuf = IdBuf<Track>;
-pub type PlaylistIdBuf = IdBuf<Playlist>;
-pub type UserIdBuf = IdBuf<User>;
-pub type ShowIdBuf = IdBuf<Show>;
-pub type EpisodeIdBuf = IdBuf<Episode>;
-
-/// A Spotify object id of given [type](crate::enums::types::Type).
+/// A Spotify object ID of a given [type](crate::enums::types::Type).
```
ramsayleung, line 140: I am a little confused that what's difference between the old Id wrapper and the new Id wrapper, it seems the only difference is that alias type declarations are moved into macro `sealed_idtypes` and an `Unknown` type has been added?

PR #242, opened 2021-08-24
Fix IDs v2

marioortizmanero:
## Description

This attempt tries to implement what i described in #203:

> A different approach would be to scratch `Id<T>` and just use `ArtistId` and similars directly. What I mean is that we wouldn't have a single type for IDs called `Id<T>` that implements its logic. We could make `ArtistId` and similars, and implement the trait the same way clients work. The main component would be a trait that implements all the logic, say `Id`, and a struct for each type (`ArtistId`, etc) that implements the trait. With this one could be generic over `Id` directly, have `Box<dyn Id> for runtime usage, and `Vec<Box<dyn Id>>` for multiple values, which is exactly what we need.
> 
>   The main problems are:
> 
>   1. We can't use type inference to our benefit; `Id::from_uri` can't work with this solution; `ArtistId::from_uri` would have to be used explicitly.
>   2. We need the trait `Id` always available when using IDs; we'd have to add it to the prelude. Considering the user needs to import the prelude to use the clients anyway I don't think it's a big deal
> 
>   So this solution is basically the same as how clients work. We have a big trait that does everything, and we let Rust "clone" the code for each of our implementations. It's also basically the newtype pattern, but with a common trait for all the newtypes.

### DIFF
--- a/rspotify-model/src/context.rs
+++ b/rspotify-model/src/context.rs
@@ -8,15 +8,15 @@ use std::time::Duration;
 
 use crate::{
     millisecond_timestamp, option_duration_ms, CurrentlyPlayingType, Device, DisallowKey, IdBuf,
-    PlayableIdType, PlayableItem, RepeatState, Type,
+    PlayableIdBuf, PlayableItem, RepeatState, Type,
 };
 
 /// Context object
 ///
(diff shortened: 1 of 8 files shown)


marioortizmanero:
Another failure, unfortunately. It's almost impossible to make these traits object-safe without putting its entire implementation in the macro :(

Issue #235, opened 2021-07-16
bug: token refresh doesn't work


insomnimus:
`rspotify::blocking::oauth2.rs line 468 called unwrap on an Err value`

This bug happens when the access token needs to be refreshed (1hour).

I've inspected the error, it seems that for some reason spotify returns this as a response to token refresh:

```json
{
	"access_token":"*************",
	"token_type":"Bearer",
	"expires_in":3600
}
```

The scope field is missing, which serde_json sees as an error since the corresponding rust struct field is not optional.

I've made it Option<String>, however then the "refreshed" token just doesn't work...

I know you are working on a new release but i'd be very glad if i could have some quick fix for this that i can apply myself, otherwise i'm just going to go write this in go (github.com/insomnimus/libman)

I tried upgrading to the master but i think the branch is just not stable yet, there were a lot of inconsistencies and i'd rather spend time writing code than migrating it.

So again, you can reproduce this error by having a client connection for ~1 hour then requesting something.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("missing field `scope`", line: 1, column: 142)', D:\home\.cargo\registry\src\github.com-1ecc6299db9ec823\rspotify-0.10.0\src\blocki
ng\oauth2.rs:468:68
```

marioortizmanero:
I strongly suggest you to use the master branch, but as you said, if you need a quick fix I did something similar myself to run some benchmarks with the old version of rspotify. I was keeping the repo private until we released v0.11 but you can take a look:

https://github.com/marioortizmanero/rspotify-bench

Let me know if `rspotify-0.10.0-patched` works, though I didn't really test the refresh tokens, only the first auth steps.

insomnimus:
Thank you for the quick reply.

I will see if your patch works for me, thanks a lot for that.

I'd love to upgrade to the master branch but it would be rather painful to rush that change on my repo, it's ~3k loc and I kind of need a quick fix right now, so eventually I'll switch to master but I don't have the energy for now.

I'll try the patch and report back if it works, thanks!

insomnimus:
Sadly, the patch does not work.

It no longer panics but after about an hour, the requests return 403 and 404, so i think the token is not refreshed properly.

I wonder if this bug is present in the master branch, i'd like to switch to it but i wanna be sure it's fixed.

Felix-Droop:
Hello, may I ask what is the current status on this? I am using the version 0.10 and also getting this error after one hour:

```
[...]

PR #244, opened 2021-08-30
Fix IDs v4

### LINKED ISSUE #203: Endpoints with lists of Ids don't accept multiple of them

marioortizmanero:
## Description

I've finally understood what the Rust compiler was trying to tell me. This PR implements type-safe ID wrappers, it is the only way I've managed to do it. Everything is thoroughly explained in `rspotify-model/src/idtypes.rs`, but if you have any questions please do let me know. In summary, I've ended up being able to make `Id` an object-safe trait, so that it can be used as `dyn Id`. This way, it is possible to take e.g. `Vec<dyn PlayableId>`.

This closes #203 

## Motivation and Context

Now that I've implemented it, the question is: are type-safe IDs worth it? There is currently a partial implementation of IDs in the `master` branch in which I'm basing this PR. We have to either finish said implementation, or remove it completely in a different PR.

Advantages:

* Type safety: with this, it is impossible for the user to make a logic error related to IDs. The user can't mistakenly pass an ID instead of an URI or vice-versa. IDs are grouped up by types now as well; one can't pass the URI of an artist instead of the URI of a track. @Felix-Droop commented in #235 that he did like strongly typed IDs, for example.
* Strongly typed IDs also makes endpoints self-documented. Taking a `TrackId` as a parameter makes it pretty clear what kind of ID is required, in comparison with `&str`, in which case it would have to be specified in the documentation (or by context).
* All IDs have useful methods available for use: `Id::uri`, `Id::url`, `Id::_type`
* The model is now more concise. Things like `FullTrack` only need an `id` field; `uri` and `_type` can be removed.

Disadvantages:

* IDs may be more cumbersome to use; there is a considerable difference between `endpoint(id)` and `endpoint(&TrackId::from_uri(id)?)`.
* Library complexity, though IDs are much easier to use than I thought initially; it's just a trait and an implementation for each type. This PR only adds about ~120 lines of code, many of which are documentation and tests.
* IDs are `String`s instead of `str`, which require an allocation. If this ends up being important in terms of performance, we could look into crates like `smol_str` (which may not allocate until the string is longer than X bytes), or even `Cow`. So performance may not be that much of a problem; we can look into it in the future.
* We lose the ability to let the compiler deduce the kind of ID we're dealing with, i.e. `Id::from_id` won't work anymore. This is only a disadvantage over the current implementation in `master`, not over not using type-safe IDs at all.

## Dependencies 

Nothing new

## Type of change

Though it is a breaking change, the only breakage over master occurs when `Id::from_id` and similars were used directly. Instead, a concrete type has to be provided now, like `ArtistId::from_id`. The other main breakage is in rspotify's model; some fields have been removed and modified so that they use this new feature instead of just strings. Additionally, `IdBuf` has been removed, since now it's always owned anyway.

## How Has This Been Tested?

There are new tests in `rspotify-model/src/idtypes.rs`. Take a look in there. The previous tests still pass.


marioortizmanero:
Something I have yet to decide: should I remove the `_type` fields in rspotify-model now that most of them have an ID with an already available type? Sounds good to me, right?

marioortizmanero:
This is ready for review, @ramsayleung. I personally like how IDs have turned out in the end; sorry for taking so long, it wasn't so complicated.

ramsayleung:
Sorry for not reviewing earlier, I took two hours to understand the context and problem came from #203. Just in case I don't get your point, I am going to explain the problem and your solution in my words.

The below endpoint takes `uris` as a parameter with the type of `impl IntoIterator<Item = &'a Id<T>> + Send + 'a`
```rust
    async fn start_uris_playback<'a, T: PlayableIdType + 'a>(
        &self,
        uris: impl IntoIterator<Item = &'a Id<T>> + Send + 'a,
        device_id: Option<&str>,
        offset: Option<crate::model::Offset<T>>,
        position_ms: Option<u32>,
    )
``` 

```rust
    // if we pass the below uris to `start_uris_playback`, the type of `uris` will be deduced to `impl IntoIterator<Item = &'a Id<Episode>>`.
    let uris = &[
    ];
	// It will fail to compile since `Track` isn't compatible with `Episode`(not the same type)
```

Your solution to fix this issue is that replacing `Id<T>` with explicit `ID` type?

marioortizmanero:
Hi Ramsay,

Sorry, the original issue is probably too confusing and lengthy to understand now. I went in circles when trying to implement these IDs and had multiple attempts until I fully understood it. So please ignore it, I'll make a quick summary:

To fully cover all the endpoints with type-safe IDs, we need to be able to have IDs of types known at runtime. This is important so that we can have a vector of any kind of ID, such as `Vec<Box<dyn Id>>`. This should also work for more specific groups of IDs like `Vec<Box<dyn PlayableId>>`. Previously, we used a generic parameter `T: PlayableId` and then `Vec<T>`, but notice that these two are far from equivalent; `Vec<Box<dyn PlayableId>>`  allows a list of episodes, tracks, or similars **all mixed up**, and `Vec<T>` allows a list of **either** episodes, tracks, or similars. Which is not what we want; [the current implementation of `start_uris_playback` is wrong](https://github.com/ramsayleung/rspotify/blob/master/src/clients/oauth.rs#L966) because you can't pass a track and an episode in the same list. The example you wrote showcases this perfectly.

Now that we know the requirements, we need a solution. In master we have `Id<T>` where `T` is the type of Id (say, `Artist`, `Album`, etc). But something like `Vec<Id<Box<dyn PlayableId>>>` is impossible, so we need to rework it. What ended up working for me was to just use dynamic dispatching. We just need an object-safe trait and then we *can* do something like `Vec<Box<dyn ObjectSafeTrait>>`. So I basically turned `Id` into an object-safe trait, and implemented it for each type of Id, resulting in `ArtistId`, `AlbumId`, etc.

Regarding your comment in #241, the thing is that in the first iteration I tried to do this in a different way without dynamic dispatching. I simply added an `Unknown` type so that we can use `Vec<Id<Unknown>>` instead, which certainly allows a list of episodes, tracks, or similars **all mixed up**. You just need to be able to freely convert e.g. `Id<Track>` into `Id<Unknown>` and then you can mix them up because they're the same type, `Unknown`. The problem is that (a) it's impossible to convert `Id<Track>` into `Id<Unknown>` and (b) when converting `Id<T>` into `Id<Unknown>` you lose information about the type of the ID, which is important to e.g. call the `uri` method (otherwise you'd get `spotify:unknown:xxxx`). This is not so important to understand because it's just a wrong approach I tried to follow. It only explains why dynamic dispatch is the only possible solution for this.

If you have any more doubts please let me know.

ramsayleung:
> But something like `Vec<Id<Box<dyn PlayableId>>>` is impossible, so we need to rework it.

Because of the requirements of zero-cost abstraction, then `Box` needs to allocate memory on heap?

> You just need to be able to freely convert Id<Track> and Id<Episode> into Id<Unknown> and then you can mix them up because they're the same type, Unknown. The problem is that (a) it's impossible to convert Id<Track> into Id<Unknown> and (b) when converting Id<T> into Id<Unknown> you lose information about the type of the ID, which is important to e.g. call the uri method

It sounds like the downcast problem in OOP language(C++, or Java), results in losing information of derived object by casting derived object to base object

CHANGELOG.md
```diff
@@ -185,15 +185,15 @@ More in the [`examples` directory](https://github.com/ramsayleung/rspotify/tree/
 - The `Spotify` client has been split up by authorization flows (`ClientCredsSpotify`, `AuthCodeSpotify`, `AuthCodePkceSpotify`), which allows us to remove the builder pattern. The authentication process has been rewritten. ([#216](https://github.com/ramsayleung/rspotify/pull/216)).
 - Fix typo in `user_playlist_remove_specific_occurrenes_of_tracks`, now it's `user_playlist_remove_specific_occurrences_of_tracks`.
 - ([#123](https://github.com/ramsayleung/rspotify/pull/123)) All fallible calls in the client return a `ClientError` rather than using `failure`.
-- ([#161](https://github.com/ramsayleung/rspotify/pull/161)) Endpoints taking `Vec<String>/&[String]` as parameter have changed to `impl IntoIterator<Item = &Id<Type>>`.
-  + The endpoints which changes parameter from `Vec<String>` to `impl IntoIterator<Item = &Id<Type>>`:
+- ([#161](https://github.com/ramsayleung/rspotify/pull/161)) Endpoints taking `Vec<String>/&[String]` as parameter have changed to `impl IntoIterator<Item = &Id>`.
```
ramsayleung, line 188: It seems #161 should be #244

rspotify-model/src/idtypes.rs
```diff
@@ -296,9 +224,243 @@ impl<T: IdType> Id<T> {
...
+                }
+
+                #[inline]
+                fn _type_static() -> Type where Self: Sized {
+                    Type::$type
+                }
+
+                #[inline]
+                unsafe fn from_id_unchecked(id: &str) -> Self {
+                    $name(id.to_owned())
```
ramsayleung, line 311: I am a little unsure that is it possible to construct a `$name`(for example `AlbumId`) from `String`? It seems there is no such constructor?

rspotify-model/src/idtypes.rs
```diff
@@ -149,99 +114,57 @@ pub enum IdError {
...
+/// The main interface for an ID.
+///
+/// # Implementation note
+///
+/// Note that for IDs to be useful their trait must be object-safe. Otherwise,
+/// it wouldn't be possible to use `Vec<Box<dyn Id>>` to take multiple kinds of
+/// IDs or just `Box<dyn Id>` in case the type wasn't known at compile time.
+/// This is why this trait includes both [`Self::_type`] and
+/// [`Self::_type_static`], and why all of the static methods use `where Self:
+/// Sized`.
```
ramsayleung, line 126: Then, after including the `Self::_type` and `Self::_type_static` for `Id`, the compiler can deduce the specific type of `Id`?

rspotify-model/src/idtypes.rs
```diff
@@ -149,99 +114,57 @@ pub enum IdError {
...
+    where
+        Self: Sized;
+
+    /// Initialize the Id without checking its validity.
+    ///
+    /// # Safety
+    ///
+    /// The string passed to this method must be made out of alphanumeric
+    /// numbers only; otherwise undefined behaviour may occur.
+    unsafe fn from_id_unchecked(id: &str) -> Self
```
ramsayleung, line 150: Sorry for not getting your point, what are the use cases of `_type`, `_type_static` and  `from_id_unchecked`, are they necessary?

CHANGELOG.md
marioortizmanero: Yup you're right

rspotify-model/src/idtypes.rs
marioortizmanero: These methods are just getters for the Id's type just like what `Id::_type` was before this PR. See the next comment as to why `_type_static` is necessary.

rspotify-model/src/idtypes.rs
marioortizmanero: Yes, it's a [tuple struct](https://learning-rust.github.io/docs/b2.structs.html#Tuple-structs) (defined as `struct $name(String)`), which you can initialize with `$name(my_string)`

rspotify-model/src/idtypes.rs
marioortizmanero: So, the thing is that `Id` is now a trait. And just like how the Rspotify clients work, its structs need to provide getters and similars so that the implementations in `Id` work ""generically"".

* `from_id_unchecked` serves as a constructor. Since the user shouldn't initialize the ID directly (they should use `from_id` and similars), it's marked as unsafe. But otherwise there would be no way to construct the Id within the trait (how else would `from_id` be implemented, for instance?). Note that we need object safety, so it uses `where Self: Sized`
* `_type` serves as a getter for the `Type`, and so does `_type_static`. The main difference is that `_type` is a method, and `_type_static` is a static function (it doesn't take `self`). This is because there are two different use-cases to get the type: in constructors, which are static, and in regular methods:
  
  * In the `from_id` constructor if we want to get the type of the ID, we need a static function that provides it. We can't use a method that takes `&self` because the type is not constructed yet. So we just call `Self::_type_static`.
  * In `uri` if we want to get the type of the ID, we need either a static function or a method. Any of these would actually work, because the type is already constructed. The problem here is that we want to be able to use this method with a `dyn Id`. Which requires that `uri` is in `Id`'s vtable. So we can't use a static function in this case, `uri` must be a method. And in order for `uri` to be a method, it can't make calls to static functions. So `Self::_type_static` won't work, we need `self._type`.

  You can see yourself how `_type` is necessary by trying to use `_type_static` in `uri` instead. You'll get a compiler error that explains it:


  ```rust
  fn uri(&self) -> String {
      format!("spotify:{}:{}", Self::_type_static(), self.id())
  }
  ```

  results in:

  ```
  error[E0277]: the size for values of type `Self` cannot be known at compilation time
     --> rspotify-model/src/idtypes.rs:159:34
      |
  159 |         format!("spotify:{}:{}", Self::_type_static(), self.id())
      |                                  ^^^^^^^^^^^^^^^^^^ doesn't have a size known at compile-time
      |
  note: required by `idtypes::Id::_type_static`
     --> rspotify-model/src/idtypes.rs:140:5
      |
  140 | /     fn _type_static() -> Type
  141 | |     where
  142 | |         Self: Sized;
      | |____________________^
  help: consider further restricting `Self`
      |
  158 |     fn uri(&self) -> String where Self: Sized {
      |                             ^^^^^^^^^^^^^^^^^
  
  For more information about this error, try `rustc --explain E0277`.
  error: could not compile `rspotify-model` due to previous error
  ```

ramsayleung: I think I (almost) understand your design, these `from_id`, `from_uri` and `from_id_or_uri` functions both are static constructor, all `Id` types can call these constructors without warring about `vtable`(of course you could make these functions as virtual function, and all `Id` types implement them, but it's clumsy).

As for `type` and `id` methods, it's necessary to make them as virtual function, while they vary between different `Id`

marioortizmanero: > I think I (almost) understand your design, these from_id, from_uri and from_id_or_uri functions both are static constructor, all Id types can call these constructors without warring about vtable(of course you could make these functions as virtual function, and all Id types implement them, but it's clumsy).

Exactly. There's no point on making these constructors virtual functions in the vtable because you couldn't use them. `dyn Id` requires an already initialized `Id`, so a virtual constructor makes no sense AFAIK. What we want is methods like `uri` to be virtual so that we can use them in the endpoints when e.g. taking `Vec<dyn Id>`.

> As for type and id methods, it's necessary to make them as virtual function, while they vary between different Id

Yup.

rspotify-model/src/playlist.rs
```diff
@@ -34,16 +32,13 @@ pub struct SimplifiedPlaylist {
     pub collaborative: bool,
     pub external_urls: HashMap<String, String>,
     pub href: String,
-    pub id: String,
+    pub id: PlaylistId,
```
ramsayleung, line 35: It seems these `Id` types have not implemented the customized serialize and deserialize function, is it ok to replace `String` with `PlaylistId`, could `serde` handle the serialization/deserialization of these `Id` type properly?

It's same to the other `Id` type.

rspotify-model/src/track.rs
```diff
@@ -104,14 +92,7 @@ pub struct SavedTrack {
 }
 
 /// Track id with specific positions track in a playlist
-pub struct TrackPositions<'id> {
-    pub id: &'id TrackId,
+pub struct TrackPositions {
+    pub id: TrackId,
     pub positions: Vec<u32>,
 }
-
```
ramsayleung, line 111: Is it ok to remove this `new` constructor function?

marioortizmanero: Yes, I don't get why it was needed in the first place if all the fields are public. You can just construct it with `TrackPostitions { id, positions }` just like how it works for every other model.

rspotify-model/src/playlist.rs
marioortizmanero: Ooh it's probably wrong now that you mention it, as the deserialization may even accept an uri on an invalid id. I'll take a look later on.

rspotify-model/src/album.rs
```diff
@@ -33,9 +32,6 @@ pub struct SimplifiedAlbum {
     pub release_date_precision: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub restrictions: Option<Restriction>,
-    #[serde(rename = "type")]
-    pub _type: Type,
-    pub uri: Option<String>,
```
ramsayleung, line 38: It seems you just removed these `_type`, `uri` and `id` fields, could they serialize/deserialize into a single `AlbumId` type?

Same for the `FullAlbum` type

marioortizmanero: My bad, I shouldn't have removed the `id` field.

rspotify-model/src/playlist.rs
marioortizmanero: This should be fixed after the last commit

rspotify-model/src/album.rs
marioortizmanero: What do you mean? That I should remove the `Option` for the `Id` field?

ramsayleung: Yes, there are `Id` fields in both of these structs, then what I mean is that there are three fields named `_type`, `uri`, `id` in the `SimplifiedAlbum` struct, could they map into one individual field named `id`?

For example, there is a `Json` data like this:

```json
// `SimplifiedAlbum` data
{
 ...
 "id": "v5eMCfY3fg3kDdBnfD80ZOW9eYybjcND",
 "_type": "album",
 "uri":"spotify_uri_xxx"
 ...
}
```

And you want to map the above data to a `AlbumId` type, while we have three input fields, but have only one output field, it's many-to-one mapping:

`id`+`_type`+`uri` => `AlbumId`

As `serde` could only handle one-to-one mapping(As far as I know)

`id` => `AlbumId`, `_type` => `AlbumId` or `uri` => `AlbumId`.

It seems impossible to implement this serialization, perhaps you could write a unit test to check if it works.

Do I demonstrate this clear? Or I just misunderstand something about how `serde` works?

ramsayleung: You take a look at the `serde` document, probably the `flatten` attribute is what you want: https://serde.rs/attr-flatten.html

marioortizmanero: The thing is that `_type` and `uri` are redundant and don't offer any more information than what we already have when deserializing. Given an ID we already know its type. So there's no need to even check these fields, we can just ignore them.

ramsayleung: I misunderstood your point, you are right.

CHANGELOG.md
```diff
@@ -185,15 +185,15 @@ More in the [`examples` directory](https://github.com/ramsayleung/rspotify/tree/
 - The `Spotify` client has been split up by authorization flows (`ClientCredsSpotify`, `AuthCodeSpotify`, `AuthCodePkceSpotify`), which allows us to remove the builder pattern. The authentication process has been rewritten. ([#216](https://github.com/ramsayleung/rspotify/pull/216)).
 - Fix typo in `user_playlist_remove_specific_occurrenes_of_tracks`, now it's `user_playlist_remove_specific_occurrences_of_tracks`.
 - ([#123](https://github.com/ramsayleung/rspotify/pull/123)) All fallible calls in the client return a `ClientError` rather than using `failure`.
-- ([#161](https://github.com/ramsayleung/rspotify/pull/161)) Endpoints taking `Vec<String>/&[String]` as parameter have changed to `impl IntoIterator<Item = &Id<Type>>`.
-  + The endpoints which changes parameter from `Vec<String>` to `impl IntoIterator<Item = &Id<Type>>`:
+- ([#244](https://github.com/ramsayleung/rspotify/pull/244)) Endpoints taking `Vec<String>/&[String]` as parameter have changed to `impl IntoIterator<Item = &Id>`.
+  + The endpoints which changes parameter from `Vec<String>` to `impl IntoIterator<Item = &Id>`:
```
ramsayleung, line 189: I think you should mention that you have removed the `uri` and `type` field for some model types in the *Break Change* part.

rspotify-model/src/idtypes.rs
```diff
@@ -296,9 +224,311 @@ impl<T: IdType> Id<T> {
...
+
+                #[inline]
+                unsafe fn from_id_unchecked(id: &str) -> Self {
+                    $name(id.to_owned())
+                }
+            }
+
+            // Deserialization may take either an Id or an URI, so its
+            // implementation has to be done manually.
+            impl<'de> Deserialize<'de> for $name {
```
ramsayleung, line 317: I am not sure if it's necessary to implement the `serialize` function manually? Otherwise, the serialized output will be different from the original `Json` data?

rspotify-model/src/show.rs
```diff
@@ -35,7 +37,6 @@ pub struct SimplifiedShow {
     pub publisher: String,
     #[serde(rename = "type")]
     pub _type: String,
```
ramsayleung, line 39: Should we remove the `_type` field?

rspotify-model/src/show.rs
```diff
@@ -136,7 +135,6 @@ pub struct FullEpisode {
     pub show: SimplifiedShow,
     #[serde(rename = "type")]
     pub _type: String,
```
ramsayleung, line 137: Same here

rspotify-model/src/show.rs
```diff
@@ -106,7 +106,6 @@ pub struct SimplifiedEpisode {
     pub resume_point: Option<ResumePoint>,
     #[serde(rename = "type")]
     pub _type: String,
```
ramsayleung, line 108: Same here

rspotify-model/src/show.rs
```diff
@@ -76,7 +77,6 @@ pub struct FullShow {
     pub publisher: String,
     #[serde(rename = "type")]
     pub _type: String,
```
ramsayleung, line 79: Same here

src/clients/oauth.rs
```diff
@@ -253,7 +253,7 @@ pub trait OAuthClient: BaseClient {
     async fn playlist_replace_tracks<'a>(
         &self,
         playlist_id: &PlaylistId,
-        track_ids: impl IntoIterator<Item = &'a TrackId> + Send + 'a,
+        track_ids: impl IntoIterator<Item = &'a dyn PlayableId> + Send + 'a,
```
ramsayleung, line 256: Since it's an endpoint to manipulate track, it doesn't need to be  `dyn PlayableId`. `TrackId` is enough, right?

CHANGELOG.md
marioortizmanero: Will do

rspotify-model/src/show.rs
marioortizmanero: Ahh the thing is that `SimplifiedShow` is undocumented, so we don't know what `_type` is exactly. I will leave it as it is to avoid breakage and add a note.

marioortizmanero: It probably works just like `FullShow`, in which case it's correct to do so. So, done.

rspotify-model/src/show.rs
marioortizmanero: Done. I've tried the endpoint and its value is `"show"`, so it can indeed be removed: https://developer.spotify.com/console/get-show/?id=38bS44xjbVVZ3No3ByF1dJ

rspotify-model/src/show.rs
marioortizmanero: Tried it in the console and it seems we can remove it, as its value is `"episode"`: https://developer.spotify.com/console/get-episode/?id=512ojhOuo1ktJprKbVcKyQ&market=

rspotify-model/src/show.rs
marioortizmanero: Same as `FullEpisode`, I guess

src/clients/oauth.rs
marioortizmanero: I've just made a commit that leaves this and similar occurrences as a TODO for another PR: https://github.com/ramsayleung/rspotify/pull/244/commits/98c299c14505c08ebe14e84061a1155f45511c3d. This is because most of the playlist endpoints work with both tracks *and* episodes, and we don't currently take that into account. I left it as a TODO to avoid making big changes so that this can be merged finally.

rspotify-model/src/idtypes.rs
marioortizmanero: The thing is that when deserializing we want to be more flexible and accept both IDs and URIs. And when serializing we can only choose one, so IDs make the most sense to me.

Though I think the problem here is that these models shouldn't even implement `Serialize`. At no point in our code do we need to serialize an ID and it's probably confusing to do so anyway. IMO we should remove the `#[derive(Serialize)]` from the IDs and many other types in the model with the same issue. I would leave that for another PR, though.

ramsayleung (approved):

ramsayleung:
I think this PR is ready to merge :)

marioortizmanero:
Great! Thanks for the review :)

ramsayleung:
Ha, thanks for your contribution, you are the guy worked hard :)

PR #245, opened 2021-08-30
Fix cached token scopes

Rigellute:
## Description

Serde rename "scopes" to "scope", but leave struct type as "scopes".

## Motivation and Context

The cached auth token was not being restored from the cache, prompting the login flow every time.

On inspecting the cache, the "scopes" property was empty.

Turns out this was because the form response from spotify does not contain a "scopes" field. But instead contains a "scope".

This PR simply renames the property to "scopes" as it's a small change. Let me know if there's a more appropriate fix.

## Type of change

## How Has This Been Tested?

- [ ] Test A: xxx
- [ ] Test B: xxx

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -320,7 +320,8 @@ pub struct Token {
     /// A list of [scopes](https://developer.spotify.com/documentation/general/guides/scopes/)
     /// which have been granted for this `access_token`
     /// You could use macro [scopes!](crate::scopes) to build it at compile time easily
-    #[serde(default, with = "space_separated_scopes")]
+    // The token response from spotify is singular, hence the rename to `scope`
+    #[serde(default, with = "space_separated_scopes", rename = "scope")]
     pub scopes: HashSet<String>,
 }
 


Rigellute:
Hello! Author of spotify-tui here.

I was playing around with the master version of rspotify (I wanted to upgrade to tokio v1), and I had trouble with auth.

Every time I started up the app, I needed run through the auth flow in the cli (copy/paste the URL). I debugged it, and this was the fix I found.

Rigellute:
Looks like [this commit](https://github.com/ramsayleung/rspotify/commit/476179988673d4c2868789596d258cf0f8002d60) might have introduced this bug.

marioortizmanero:
Oh damn we should've caught that with tests. I made that commit because we were mixing "scope" and "scopes" throughout the crate and I thought it was confusing, so I started using "scopes" consistently, which is also how the API docs refer to it (take a look [here](https://developer.spotify.com/documentation/general/guides/authorization-guide/)).

So your solution is definitely the way to go, thanks for the PR. Before merging it I'll try to write a couple tests to make sure everything works properly.

marioortizmanero:
Ah, right. We didn't catch it with tests because it only happens for oauth clients and we can't currently run tests with that. Still, I'll make a test for it.

marioortizmanero (approved):
I've made sure it works myself. Unfortunately it's not easy to make a test for it, so I'll just merge this for now. Thanks for the PR!

PR #249, opened 2021-08-31
Simplify recommendations endpoint

marioortizmanero:
## Description

The recommendations endpoint was very messy and had a long-standing TODO item. I've simplified it so that it just takes a map with the attributes that are directly appended to the query payload.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

The `test_recommendations` endpoint

### DIFF
--- a/src/clients/base.rs
+++ b/src/clients/base.rs
@@ -887,6 +887,9 @@ where
     /// Get Recommendations Based on Seeds
     ///
     /// Parameters:
+    /// - attributes - restrictions on attributes for the selected tracks, such
+    ///   as `min_acousticness` or `target_duration_ms`. See the reference for a
+    ///   list of all of them.
     /// - seed_artists - a list of artist IDs, URIs or URLs
     /// - seed_tracks - a list of artist IDs, URIs or URLs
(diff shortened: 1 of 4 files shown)


src/clients/base.rs
```diff
@@ -901,7 +906,7 @@ where
     /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-recommendations)
     async fn recommendations<'a>(
         &self,
-        payload: &Map<String, Value>,
+        attributes: impl IntoIterator<Item = (&'a str, Value)> + Send + 'a,
```
ramsayleung, line 909: Is it better to make attribute as a named type(such as `Enum`) instead of a string?

marioortizmanero: Done. I wanted to avoid the huge enum but it's indeed much easier to use this way.

ramsayleung:
LGTM, since you have changed the signature of this endpoint, would you like to update the **break change** part of `CHANGELOG.md`

marioortizmanero:
Sure!

ramsayleung (approved):

PR #250, opened 2021-09-06
Cleanup

marioortizmanero:
## Description

This cleans up quite a few things:

* Removes unnecessary dependencies
* Uses `debug_assert` instead of `log` to ensure endpoint predicates are met; some people don't enable logging at all
* Clean up unnecessary `derive` macros.
* Move all `Serialize/Deserialize` items from `src` to `rspotify-model`, like `Token`, where they fit better. This forced me to create a `ModelError`, but I don't really see the problem.
* Move custom [de]serialization modules into `custom_serde` and fix `pub(in crate)` occurences
* `Token::from_cache` now returns a `Result` instead of an `Option`, as the TODO above indicated.
* Bumps all the dependencies

## Motivation and Context

I'm working on the last things needed to finally release v0.11. This is one of them :)

I got some of these suggestions from this very nice article, give it a read yourself: https://matklad.github.io/2021/09/04/fast-rust-builds.html

## Dependencies 

None, if anything I've removed some.

## Type of change

## How Has This Been Tested?

The CI should also pass after this


marioortizmanero:
I've ran some compilation profiling with the following command:

```
cargo +nightly build -Z timings --release
```

And got these results:

[cargo-timing.zip](https://github.com/ramsayleung/rspotify/files/7115269/cargo-timing.zip)

marioortizmanero:
I've updated the dependencies with `cargo outdated --workspace --depth=1`. But I haven't seen dependabot in a long time. Is it still working, @ramsayleung ?

marioortizmanero:
Here are the results of this PR, run them yourself as well if you want to make sure:

* **Before**: 
  * reqwest: 159 compilation units, 1m 2s
  * ureq: 120 compilation units, 55s
* **After**: 
  * reqwest: 159 compilation units, 1m 21s
  * ureq: 102 compilation units, 52s

So in the end the cleanup does help a tiny bit in terms of compilation time, specially in the case of ureq. The `Cargo.toml` and similars are much tidier, too.

ramsayleung:
> I got some of these suggestions from this very nice article, give it a read yourself: matklad.github.io/2021/09/04/fast-rust-builds.html

Yes, I have seen this post in the [`/r/rust`](https://www.reddit.com/r/rust/), and took a rough glance to this article. I would like to read it later.

 > But I haven't seen dependabot in a long time. Is it still working, @ramsayleung ?

It should work, I haven't changed anything of this bot.

> So in the end the cleanup does help a bit in terms of compilation time

It seems the compilation time of debug build and release build for `reqwest` both grow a bit.

rspotify-http/Cargo.toml
```diff
@@ -18,20 +18,15 @@ rspotify-model = { path = "../rspotify-model", version = "0.10.0" }
...
-futures = { version = "0.3.8", optional = true }
-log = "0.4.11"
-maybe-async = "0.2.4"
-reqwest = { version = "0.11.0", default-features = false, features = ["json", "socks"], optional = true }
-serde_json = "1.0.57"
-thiserror = "1.0.20"
-ureq = { version = "2.0", default-features = false, features = ["json", "cookies"], optional = true }
-url = "2.2.2"
+async-trait = { version = "0.1.51", optional = true }
+log = "0.4.14"
```
ramsayleung, line 22: Since you have replaced `log` with `debug_assert`, is it possible to remove this crate?

ramsayleung (approved):

rspotify-http/Cargo.toml
marioortizmanero: I was going to remove it but I think it's still useful to have around. I wanted to add some log statements before releasing the new version because we don't even log that much right now. But I wanted to wait for the pending PRs to be merged first.

marioortizmanero:
Also, nah I wouldn't introduce bors. It seems complicated and we're a small team so not really necessary.

Reqwest might be a bit slower but that's just the benchmark innaccuracy. I only did one run and the differences are very small so I would treat it as a non-improvement but also a non-downgrade. It's more or less the same.

ramsayleung:
Is it possible to benchmark with CI, just treating every check PR as build benchmark. Thus, we could check the last CI build as the `before` benchmark and current build as `after` benchmark?

marioortizmanero:
I mean you can always take a look at how long it takes the CI to finish, but it's just as inaccurate as doing it manually because there's only one run.

* Last commit in this PR:
  * https://github.com/ramsayleung/rspotify/runs/3524321817 ureq: 2m 49s
  * https://github.com/ramsayleung/rspotify/runs/3524321899 reqwest: 4m 4s
* Last commit in master:
  * https://github.com/ramsayleung/rspotify/runs/3531367936 ureq: 2m 49s
  * https://github.com/ramsayleung/rspotify/runs/3531367978 reqwest: 4m 35s

The thing is that even if the changes in this PR don't improve compile times, it's still an improvement because it tidies up the Cargo.toml files and similars.

ramsayleung:
Would you like to remove the `log` crate, otherwise I think this PR is ready to merge?

marioortizmanero:
I'll add some more logs to the main crate after the rest of the PRs are merged, so we'll need it. I'll merge then.

PR #251, opened 2021-09-15
Allow negative values in duration_ms

flip1995:
## Description

This patch adds a parser for `i64` values to `duration_ms`, so that it no longer fails for negative numbers. This works, according to the [serde documentation](https://serde.rs/impl-deserialize.html#driving-a-visitor):

> The JSON `Deserializer` will call `visit_i64` for any signed integer and `visit_u64` for any unsigned integer, even if hinted a different type.

## Motivation and Context

The spotify API doesn't make any guarantees, if the returned integers are signed or unsigned. It turns out, that for the `resume_position_ms` value in the `ResumePointObject` used for `Episode`s can be negative. I've only seen the value `-1000`, but don't know when this occurs or if other values are possible.

I noticed this using the `spotify-tui` crate, that uses this crate as a dependency. See screenshot below:
image

## Dependencies 

None

## Type of change

## How Has This Been Tested?

By adding a unit test to the `test_models` file. The previous behavior is unchanged (according to existing tests)

### DIFF
--- a/rspotify-model/src/custom_serde.rs
+++ b/rspotify-model/src/custom_serde.rs
@@ -18,6 +18,12 @@ pub mod duration_ms {
         {
             Ok(Duration::from_millis(v))
         }
+        fn visit_i64<E>(self, v: i64) -> Result<Self::Value, E>
+        where
+            E: de::Error,
+        {
+            Ok(Duration::from_millis(v.max(0) as u64))
+        }
     }
 
     /// Deserialize `std::time::Duration` from milliseconds (represented as u64)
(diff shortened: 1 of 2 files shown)


ramsayleung (approved):

ramsayleung:
Thanks for your contribution, merged. I am not sure if it's a bug of Spotify API, according to the document:
> resume_position_ms: The user’s most recent position in the episode in milliseconds. 

I don't know why would this be negative values, perhaps you could report a bug to Spotify.

flip1995:
It might be a bug in the API. Just after opening this PR I encountered the same bug with the Player API and synchronizing between devices.¹ So this might happen across APIs that has something to do with timings, so I'd say this is generally a good fix.

PR #252, opened 2021-09-19
Fixes cached tokens

marioortizmanero:
## Description

This fixes:

*  `prompt_for_user`, which failed when `read_token_cache` returned and error. The problem here is that we *are* expecting an error in case the cache file isn't found (`FileNotFoundError`). We just want to ignore it, but it used `?`, which caused an early return. So the `match` expression just accepts `Ok(Some())` now.
* `read_token_cache` for OAuth clients didn't check the config. Whoops

I also updated the docs to make them more exhaustive.

## Motivation and Context

This was caused by #250 

## Dependencies 

None

## Type of change

## How Has This Been Tested?

If you remove all cached tokens in your main directory and try to run any example it won't work. Now it does. This is because `prompt_user_token` returned `FileNotFound` and propagated. For example try with:

```
cargo run --example auth_code --features="env-file,cli"
```

### DIFF
--- a/src/clients/oauth.rs
+++ b/src/clients/oauth.rs
@@ -38,16 +38,25 @@ pub trait OAuthClient: BaseClient {
     /// token will be saved internally.
     async fn refresh_token(&mut self, refresh_token: &str) -> ClientResult<()>;
 
-    /// Tries to read the cache file's token, which may not exist.
+    /// Tries to read the cache file's token.
+    ///
+    /// This will return an error if the token couldn't be read (e.g. it's not
+    /// available or the JSON is malformed). It may return `Ok(None)` if:
+    ///
+    /// * The read token is expired
+    /// * The cached token is disabled in the config
     async fn read_token_cache(&mut self) -> ClientResult<Option<Token>> {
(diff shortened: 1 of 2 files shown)


src/clients/oauth.rs
```diff
@@ -38,16 +38,25 @@ pub trait OAuthClient: BaseClient {
...
     async fn refresh_token(&mut self, refresh_token: &str) -> ClientResult<()>;
 
-    /// Tries to read the cache file's token, which may not exist.
+    /// Tries to read the cache file's token.
+    ///
+    /// This will return an error if the token couldn't be read (e.g. it's not
+    /// available or the JSON is malformed). It may return `Ok(None)` if:
+    ///
+    /// * The read token is expired
+    /// * The cached token is disabled in the config
```
ramsayleung, line 47: It seems you are missing one condition that
```rust
/// * Its scopes don't match with the current client
```

marioortizmanero: Whoops, I messed up the docs. Should be fine now

ramsayleung (approved):

PR #253, opened 2021-09-19
Fix feature conflicts v2

### LINKED ISSUE #221: Avoiding conflicting features; allow multiple HTTP clients at once

marioortizmanero:
## Description

This supersedes #222, which is now kinda outdated.

In this attempt I'm trying to follow what the https://github.com/bodil/im-rs crate did. We can just have two crates; the main one for async, and a separate one for sync. We can keep using `maybe_async` in theory. This is unfortunately a bit confusing but it's the only way to do it. I'm working on a blog post with research about the whole async & sync stuff and I'll post it before this PR is merged to give more context and a full explanation of how it works.

## Motivation and Context

This closes #221 

## Dependencies 

None

## Type of change

## How Has This Been Tested?

An empty project with this `Cargo.toml` will fail:

```toml
[package]
name = "conflict"
version = "0.1.0"
edition = "2018"

[dependencies]
# Method number one
rspotify = { git = "https://github.com/ramsayleung/rspotify.git", branch = "master", features = ["client-reqwest", "client-ureq"]}

# Method number two
# rspotify-sync = { git = "https://github.com/ramsayleung/rspotify.git", branch = "fix-conflicts", features = ["client-reqwest", "client-ureq"] }
# rspotify-async = { git = "https://github.com/ramsayleung/rspotify.git", branch = "fix-conflicts", features = ["client-reqwest", "client-ureq"] }
```

But with this it will work:

```toml
[package]
name = "conflict"
version = "0.1.0"
edition = "2018"

[dependencies]
# Method number one
# rspotify = { git = "https://github.com/ramsayleung/rspotify.git", branch = "master", features = ["client-reqwest", "client-ureq"]}

# Method number two
rspotify-sync = { git = "https://github.com/ramsayleung/rspotify.git", branch = "fix-conflicts-2", features = ["client-ureq"] }
rspotify-async = { git = "https://github.com/ramsayleung/rspotify.git", branch = "fix-conflicts-2", features = ["client-reqwest"] }

```


marioortizmanero:
Nah, I haven't managed to get this working either... Even though it does compile individually, a project with both `rspotify-async` and `rspotify-sync` will fail to compile. Even with different `rev` values (meaning that the crates shouldn't really be merged, because they are from different sources).

PR #254, opened 2021-09-20
Fix oauth set token and fix unauthorized playback control

Rigellute:
## Description

This PR fixes two issues:

1. The `prompt_for_token` oauth flow was not loading the token after reading from the cache.
2. All calls to `start_context_playback` resulted in "request unauthorized"

## Motivation and Context

Regarding the first issue:  Looks like the previous code was expecting the `String.replace()` method to mutate the original string, but `replace` [creates a new string](https://doc.rust-lang.org/std/string/struct.String.html#method.replace). I've added a new method to explicitly set the token. Unless there's a more consitent way?

For the second issue, the `start_context_playback` method was using the `.put` method and passing `None` for the headers, hence why all requests were unauthorized. I switched this to use `endpoint_put`, which appears to simply abstract the headers placement?

## Dependencies 

None.

## Type of change

## How Has This Been Tested?

For the token cache issue, you can reproduce on master by 

1. setting the auth config `token_cached: true`
2. run prompt_for_token the first time to get the token (all requests are auth'ned)
3. restart the program to load the token from the cache
4. all requests are not auth'ned

For the not authorized playback control issue:

1. Get the auth token and device id and a playback context uri
2. Make the request and see the error

### DIFF
--- a/src/auth_code_pkce.rs
+++ b/src/auth_code_pkce.rs
@@ -43,8 +43,8 @@ impl BaseClient for AuthCodePkceSpotify {
         self.token.as_ref()
     }
 
-    fn get_token_mut(&mut self) -> Option<&mut Token> {
-        self.token.as_mut()
+    fn get_token_mut(&mut self) -> &mut Option<Token> {
+        &mut self.token
     }
 
     fn get_creds(&self) -> &Credentials {
--- a/src/auth_code_pkce.rs
+++ b/src/auth_code_pkce.rs
@@ -64,6 +64,10 @@ impl OAuthClient for AuthCodePkceSpotify {
         &self.oauth
     }
 
+    fn set_token(&mut self, token: Option<Token>) {
+        self.token = token;
+    }
+
     async fn request_token(&mut self, code: &str) -> ClientResult<()> {
         // TODO
         let mut data = Form::new();
--- a/src/auth_code_pkce.rs
+++ b/src/auth_code_pkce.rs
@@ -64,10 +64,6 @@ impl OAuthClient for AuthCodePkceSpotify {
         &self.oauth
     }
 
-    fn set_token(&mut self, token: Option<Token>) {
-        self.token = token;
-    }
-
     async fn request_token(&mut self, code: &str) -> ClientResult<()> {
         // TODO
         let mut data = Form::new();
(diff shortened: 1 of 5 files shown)


Rigellute:
Forgive me for fixing both issues in one PR, I forgot I fixed the first issue a while back in my fork. Hope it's okay to include here, as it's small. But let me know if you want both fixes in individual PRs.

src/auth_code_pkce.rs
```diff
@@ -64,6 +64,10 @@ impl OAuthClient for AuthCodePkceSpotify {
         &self.oauth
     }
 
+    fn set_token(&mut self, token: Option<Token>) {
```
marioortizmanero, line 67: Why not use `get_token_mut`?

src/clients/oauth.rs
```diff
@@ -969,7 +972,7 @@ pub trait OAuthClient: BaseClient {
         };
 
         let url = append_device_id("me/player/play", device_id);
-        self.put(&url, None, &params).await?;
+        self.endpoint_put(&url, &params).await?;
```
marioortizmanero, line 975: Good catch on that one, thanks for the fix

marioortizmanero:
I think it would be better to keep using `get_token_mut` but without `replace`. Can you just do `self.get_token_mut() = token` or similars? Otherwise `get_token_mut` is rather pointless.

marioortizmanero:
I figured it out by taking a look at how `HashMap` does it (https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.get_mut)

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0d47e91d64d301f8cc4d8f98dedbea05

Rigellute:
@marioortizmanero ah nice! But now what would happend if `get_token_mut` returns `None`? We wouldn't set it due to the `if let Some(..)`?

marioortizmanero:
Ah dang you're right. We should set it when it's None as well. In that case we should make `get_token_mut` return `&mut Option<T>` instead of `Option<&mut T>`. Otherwise we can't modify the `Option` itself.

Here's the playground link:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=c394520548d66e6514ccb9fdd8f41071

Rigellute:
Ah of course! Nice one 👏. Okay great, I've updated the method signature in the trait and update the impls.

marioortizmanero (approved):
LGTM!

PR #255, opened 2021-09-25
Fix refreshing token

marioortizmanero:
## Description

This fixes #224 by creating a wrapper for `Mutex` so that it's the same for both sync and async programming, similarly to how pagination works.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,29 @@
+on:
+  create:
+    tags:
+      - v*
+
+name: Publish Crate
+
+jobs:
+  publish:
+    name: Deploy a new version of Rspotify to crate.io
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - name: Release
+        env:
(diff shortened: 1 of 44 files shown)


marioortizmanero:
Ok so I've got it compiling without needing lots of `#[cfg()]`. The only problem is that there are probably lots of deadlocks, I should try to fix these before this is merged.

marioortizmanero:
What do you think about renaming `get_token` to `reauth_token`, @ramsayleung? I remember you didn't like `get_token` because the `.await` was weird. Perhaps `reauth_token` would be better? Maybe `get_or_reauth_token` otherwise.

marioortizmanero:
While I'm at it I've also merged master onto this branch and fixed all the conflicts

marioortizmanero:
Damnit, I've completely messed up the diff by merging master...

marioortizmanero:
Well, this has become a mess. I'm having lots of trouble merging master... Would you mind re-doing the PR with the current `master` branch, @ramsayleung? It's probably going to be much easier than trying to merge with this super outdated branch.

It shouldn't be too hard, you can copy-paste most stuff (the examples and the `mutex` module, for one)

ramsayleung:
Since everything else has been merged, we could continue work on #224 :)

ramsayleung:
Did you run the `with_auto_reauth` example successfully? I got an `stackoverflow` error when I ran it.

marioortizmanero:
That most likely means that there's unbounded recursivity somewhere, i. e. the auto reauth function calls something which calls the auto reauth again. I'll take a look later.

PR #256, opened 2021-09-25
 Added missing parameter from current_user_recently_played()

marioortizmanero:
Credits to @hellbound22 for the original PR: #198. I've rewritten it for the updated `rspotify` and applied the review comments.

## Description

Fix the implementation of ```current_user_recently_played()``` with the missing parameters.

## Motivation and Context

The Spotify web api of ```current_user_recently_played()``` has the parameters ```before``` and ```after``` that are not implemented in this crate.

## Dependencies 

None

## Type of change

## How Has This Been Tested?

The OAuth2 test passes when adding the ```None``` as the second parameter. Might have to write another test when the Option is ```Some(_)```

Also, been using this with success in a project of mine.

- [x] ```test_with_oauth::test_current_user_recently_played()```

### DIFF
--- a/src/clients/oauth.rs
+++ b/src/clients/oauth.rs
@@ -701,16 +701,32 @@ pub trait OAuthClient: BaseClient {
     ///
     /// Parameters:
     /// - limit - the number of entities to return
+    /// - time_limit - a Unix timestamp in milliseconds. Returns all items after
+    /// or before (but not including) this cursor position. 
     ///
-    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-the-users-currently-playing-track)
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#endpoint-get-recently-played)
     async fn current_user_recently_played(
         &self,
         limit: Option<u32>,
+        time_limit: Option<TimeLimits>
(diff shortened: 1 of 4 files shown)


src/clients/oauth.rs
```diff
@@ -701,17 +701,29 @@ pub trait OAuthClient: BaseClient {
...
     ) -> ClientResult<CursorBasedPage<PlayHistory>> {
         let limit = limit.map(|x| x.to_string());
-        let params = build_map! {
+        let mut params = build_map! {
             optional "limit": limit.as_deref(),
         };
 
+        let time_limit = match time_limit {
+            Some(TimeLimits::Before(y)) => Some(("before", y.timestamp().to_string())),
+            Some(TimeLimits::After(y)) => Some(("after", y.timestamp().to_string())),
```
ramsayleung, line 720: According to the document:
> `after`: A Unix timestamp in **milliseconds**
> `before`: A Unix timestamp in **milliseconds**

Then `DateTime:: timestamp` returns the number of non-leap **seconds** since January 1, 1970 0:00:00 UTC (aka "UNIX timestamp"). You should get the `milliseconds` value instead of `seconds` value.

marioortizmanero: Good catch! My bad

ramsayleung (approved):

Issue #137, opened 2020-10-11
Clean up and re-structure the errors


marioortizmanero:
The `ClientError` type currently defined for the next release could definitely be improved. Some of its variants are repetitive and confusing. I kept adding variants for the rewrite without thinking much about the structure because the important thing was to get the rewrite done. But I would like to discuss what it should be like.

Here's more or less the current definition:

```rust
#[derive(Debug, Error)]
pub enum ClientError {
    #[error("invalid client authentication: {0}")]
    InvalidAuth(String),

    #[error("request unauthorized")]
    Unauthorized,

    #[error("exceeded request limit")]
    RateLimited(Option<usize>),

    #[error("request error: {0}")]
    Request(String),

    #[error("status code {0}: {1}")]
    StatusCode(u16, String),

    #[error("spotify error: {0}")]
    API(#[from] APIError),

    #[error("json parse error: {0}")]
    ParseJSON(#[from] serde_json::Error),

    #[error("url parse error: {0}")]
    ParseURL(#[from] url::ParseError),

    #[error("input/output error: {0}")]
    IO(#[from] std::io::Error),

    #[cfg(feature = "cli")]
    #[error("cli error: {0}")]
    CLI(String),

    #[error("cache file error: {0}")]
    CacheFile(String),
}
```

Some thoughts: 
* `StatusCode`, `Request`, `RateLimited` and `Unauthorized` could be wrapped under the same variant. These are just possible HTTP errors, so something like `HTTPError` could do fine. I understand that having `Unauthorized` and such can be useful to match against these specific errors, but if `HTTPError` was defined properly it wouldn't be necessary (say, using [`http::status::StatusCode`](https://docs.rs/http/0.2.1/http/status/struct.StatusCode.html), for example).
* Maybe `CLI` should be renamed to `UserError`? It describes better its purpose.
* Is `CacheFile` necessary? I don't see it used anywhere, and it could be replaced with `IO`.
* Should `ParseURL` and `ParseJSON` be wrapped under a `Parse` variant? Maybe that's unnecessary?
* Should `InvalidAuth` be renamed to something different? It might be confusing when compared to `Unauthorized`, but `InvalidAuth` means that Spotify wasn't initialized correctly (say, there's no token when it's needed). Maybe `MissingAuth`? IDK
* None of these errors are documented (with `///`). They should include a comment with a brief description.

Please let me know any other problems you see and if you'd like to change something else.

hrkfdn:
Hey,

[...]

PR #258, opened 2021-09-25
HTTP Error cleanup

marioortizmanero:
## Description

This cleans up the errors for `rspotify-http`, according to #137.

## Motivation and Context

The [current error type for http](https://github.com/ramsayleung/rspotify/blob/bcfb7842aae7d2a0b8c82c53966efacc7e1f3966/rspotify-http/src/common.rs#L13) tries to wrap up all the errors from the clients we support under the same interface. However, it's a mess because:

1. It mixes multiple kinds of errors for different clients; `HttpError::Io` is only needed for ureq
2. It serializes only specific kinds of errors. Why is there a variant for `RateLimited` or `Unauthorized` but it's `StatusCode` for the rest?
3. Aren't the `Api` and `StatusCode` variants the same?

So this PR attempts to fix it by simply having a variant of error for each client supported by `rspotify-http`. It's as easy as delegating error handling to the HTTP client. If the user has enabled `ureq`, then they should deal with `ureq::Error` instead of some wrapper of our own that might not have everything they need.

This way, error handling is *complete*. There is nothing of functionality we miss by wrapping all of them under the same interface. And we also don't have to maintain such a complex wrapper ourselves.

I remember that @hrkfdn pointed out that the `RateLimited` variant was a must for him in order to retry a request. Well, this design still makes it possible to do so. It's just dependent on what HTTP client he's using. If he uses reqwest then it should be something like:

```rust
match request(...) {
    ClientError::Http(HttpError::Reqwest(ReqwestError::StatusCode(response))) => {
        // This is a reqwest::Response with full information about the error
        let timeout = response
            .headers()
            .get(reqwest::header::RETRY_AFTER)
            .and_then(|header| header.to_str().ok())
            .and_then(|duration| duration.parse().ok());
        time::sleep(timeout);
    }
    _ => { todo!() }
}
```

And if someone wanted to get the API error from Spotify:

```rust
match request(...) {
    ClientError::Http(HttpError::Reqwest(ReqwestError::StatusCode(response))) => {
        let api_err = response
            .json::<rspotify::model::ApiError>()?;
    }
    _ => { todo!() }
}
```

## Dependencies 

None

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?

The CI still passes

### DIFF
--- a/rspotify-http/src/reqwest.rs
+++ b/rspotify-http/src/reqwest.rs
@@ -9,7 +9,16 @@ use maybe_async::async_impl;
 use reqwest::{Method, RequestBuilder};
 use serde_json::Value;
 
-pub type Error = reqwest::Error;
+/// Custom enum that contains all the possible errors that may occur when using
+/// `reqwest`.
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("request: {0}")]
+    Client(#[from] reqwest::Error),
+
+    #[error("status code {}", reqwest::Response::status(.0))]
+    StatusCode(reqwest::Response),
+}
 
 #[derive(Default, Debug, Clone)]
 pub struct ReqwestClient {
@@ -48,16 +57,20 @@ impl ReqwestClient {
 
         // Finally performing the request and handling the response
         log::info!("Making request {:?}", request);
-        let response = request.send().await?;
+        let response = request
+            .send()
+            .await
+            .map_err(|error| HttpError::Reqwest(Error::Client(error)))?;
 
-        if response.status().is_success() {
-            response.text().await.map_err(Into::into)
-        } else {
-            Err(HttpError::StatusCode(
-                response.status().as_u16(),
-                response.json().await.ok(),
-            ))
+        // Making sure that the status code is OK
+        if !response.status().is_success() {
+            return Err(HttpError::Reqwest(Error::StatusCode(response)));
         }
+
+        response
+            .text()
+            .await
+            .map_err(|error| HttpError::Reqwest(Error::Client(error)))
     }
 }
 
--- a/rspotify-http/src/reqwest.rs
+++ b/rspotify-http/src/reqwest.rs
@@ -6,45 +6,10 @@ use super::{BaseHttpClient, Form, Headers, HttpError, HttpResult, Query};
 use std::convert::TryInto;
 
 use maybe_async::async_impl;
-use reqwest::{Method, RequestBuilder, StatusCode};
-use rspotify_model::ApiError;
+use reqwest::{Method, RequestBuilder};
 use serde_json::Value;
 
-impl HttpError {
-    pub async fn from_response(response: reqwest::Response) -> Self {
-        match response.status() {
-            StatusCode::UNAUTHORIZED => Self::Unauthorized,
-            StatusCode::TOO_MANY_REQUESTS => Self::RateLimited(
-                response
-                    .headers()
-                    .get(reqwest::header::RETRY_AFTER)
-                    .and_then(|header| header.to_str().ok())
-                    .and_then(|duration| duration.parse().ok()),
-            ),
-            status @ StatusCode::FORBIDDEN | status @ StatusCode::NOT_FOUND => response
-                .json::<ApiError>()
-                .await
-                .map(Into::into)
-                .unwrap_or_else(|_| status.into()),
-            status => status.into(),
-        }
-    }
-}
-
-impl From<reqwest::Error> for HttpError {
-    fn from(err: reqwest::Error) -> Self {
-        Self::Request(err.to_string())
-    }
-}
-
-impl From<reqwest::StatusCode> for HttpError {
-    fn from(code: reqwest::StatusCode) -> Self {
-        Self::StatusCode(
-            code.as_u16(),
-            code.canonical_reason().unwrap_or("unknown").to_string(),
-        )
-    }
-}
+pub type Error = reqwest::Error;
 
 #[derive(Default, Debug, Clone)]
 pub struct ReqwestClient {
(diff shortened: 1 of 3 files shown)


rspotify-http/src/reqwest.rs
```diff
@@ -1,49 +1,23 @@
...
-    fn from(err: reqwest::Error) -> Self {
-        Self::Request(err.to_string())
-    }
-}
+/// Custom enum that contains all the possible errors that may occur when using
+/// `reqwest`.
+#[derive(thiserror::Error, Debug)]
+pub enum ReqwestError {
+    #[error("request: {0}")]
+    Client(#[from] reqwest::Error),
```
ramsayleung, line 17: It's a little confusing that it seems `Client` and `StatusCode` are not under the same abstract level, is it fine to  wrap them into the same variant? And what does `Client` mean? What is the use case for `Client`?
Then, `StatusCode` doesn't sound like an error?

rspotify-http/src/reqwest.rs
```diff
@@ -85,23 +59,26 @@ impl ReqwestClient {
         log::info!("Making request {:?}", request);
         let response = request.send().await?;
 
+        // Making sure that the status code is OK
         if response.status().is_success() {
             response.text().await.map_err(Into::into)
         } else {
-            Err(HttpError::from_response(response).await)
+            Err(ReqwestError::StatusCode(response))
```
ramsayleung, line 66: It seems you don't show how to construct the `ReqwestError::Client()`? When will then client return this error type?

rspotify-http/src/ureq.rs
```diff
@@ -1,15 +1,25 @@
...
+/// `ureq`.
+#[derive(thiserror::Error, Debug)]
+pub enum UreqError {
+    #[error("transport: {0}")]
+    Transport(#[from] ureq::Transport),
+
+    #[error("I/O: {0}")]
+    Io(#[from] io::Error),
+
+    #[error("status code {}", ureq::Response::status(.0))]
```
ramsayleung, line 21: Would you like to add `///` doc to include a comment with a brief description?

rspotify-http/src/ureq.rs
```diff
@@ -41,21 +51,28 @@ impl UreqClient {
...
+    type Error = UreqError;
+
     #[inline]
-    fn get(&self, url: &str, headers: Option<&Headers>, payload: &Query) -> HttpResult<String> {
+    fn get(
+        &self,
+        url: &str,
+        headers: Option<&Headers>,
+        payload: &Query,
+    ) -> Result<String, UreqError> {
```
ramsayleung, line 75: You could simply return `-> Result<String, Self::Error>` as what `reqwest` did.

rspotify-http/src/ureq.rs
```diff
@@ -41,21 +51,28 @@ impl UreqClient {
...
         match send_request(request) {
-            // Successful request
             Ok(response) => response.into_string().map_err(Into::into),
-            // HTTP status error
-            Err(ureq::Error::Status(_, response)) => Err(HttpError::from_response(response)),
-            // Some kind of IO/transport error
-            Err(err) => Err(HttpError::Request(err.to_string())),
+            Err(err) => match err {
+                ureq::Error::Status(_, response) => Err(UreqError::StatusCode(response)),
+                ureq::Error::Transport(transport) => Err(UreqError::Transport(transport)),
```
ramsayleung, line 59: It seems the construction of `UreqError` is not exhaustive, we miss the `UreqError::Io` error type?

src/lib.rs
```diff
@@ -197,8 +197,11 @@ pub enum ClientError {
     #[error("url parse error: {0}")]
     ParseUrl(#[from] url::ParseError),
```
ramsayleung, line 198: Would you like to wrap `ParseJson` and `ParseUrl` into a wrapper type named `ParseError` as you mentioned in #137 ?

marioortizmanero: I honestly don't think it's necessary anymore. It will just make pattern matching more complicated

rspotify-http/src/ureq.rs
marioortizmanero: It's actually used in line 46 with Into::into

rspotify-http/src/ureq.rs
marioortizmanero: Yup you're right.

rspotify-http/src/reqwest.rs
marioortizmanero: It's just the same as `reqwest::Error`, so whenever that happens. For example when making the request or when converting the reply into a String.

rspotify-http/src/ureq.rs
marioortizmanero: Will do

rspotify-http/src/reqwest.rs
marioortizmanero: The thing is that most HTTP libraries have two kinds of errors when making a request:

* An error in the client: when making the request (for example there's no internet or a DNS failure occurs), converting the response into UTF-8, etc
* An error in the server you made a request to: a 404, 503, 403, etc. Those codes considered as successful will not be included here (< 400).

It is fine to wrap them up under the same enum because both are errors after all.

marioortizmanero:
Let me know if you like the updated documentation! This can be merged if you're ok with it.

rspotify-http/Cargo.toml
```diff
@@ -14,6 +14,8 @@ keywords = ["spotify", "api"]
 edition = "2018"
 
 [dependencies]
+rspotify-model = { path = "../rspotify-model", version = "0.10.0" }
```
ramsayleung: Is it necessary to include `rspotify-model`?

marioortizmanero: Yes, because it's used in examples in the documentation

ramsayleung:
The CI flow failed :)

PR #263, opened 2021-10-10
Refresh fixes

marioortizmanero:
## Description

Some fixes for #262:

* Improved threading example
* New threading example for async with tasks
* Fix and tidy up the `examples/with_auto_reauth.rs` example
* Moves `auto_reauth` and `refresh_token` to the trait, since we already have `refetch_token` available.
* Avoids unbounded recursivity (stack overflow)
* Simplifies everything a lot by only refreshing the token automatically when the request is being made, and not whenever it's accessed to. Thus, `auto_reauth` is only called once, and we can avoid having two token getters (one that refreshes and another that doesn't).
* `Token::is_expired` now includes a margin of 10 seconds
* `Token::can_reauth` was wrong: it checked if `refresh_token` was `Some()`, but that's not true in the case of client credentials.

## Motivation and Context

See #262

## Dependencies 

None

## Type of change

## How Has This Been Tested?

You can run the new test with:

```
cargo run --example tasks --features=cli,env-file
```

### DIFF
--- a/examples/with_auto_reauth.rs
+++ b/examples/with_auto_reauth.rs
@@ -49,7 +49,7 @@ async fn client_creds_do_things(spotify: &ClientCredsSpotify) {
 }
 
 async fn expire_token<S: BaseClient>(spotify: &S) {
-    let token_mutex = spotify.get_token().await;
+    let token_mutex = spotify.get_token();
     let mut token = token_mutex.lock().await.unwrap();
     assert!(token.is_some());
     token.as_mut().map(|x| {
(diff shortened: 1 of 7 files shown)


AndreasBackx:
Just wanted to say thanks for continuing this work! Just updated a small project of mine and I'm so happy caching and auto refresh works (haven't tried the auto refresh just yet, but caching does work). :heart: Noticed the stack overflow issue and then I saw you fixed it 5 minutes before I saw it. 😅

ramsayleung:
> No, the thing is that in should_reauth the lock is dropped after turning it into a boolean because it's a single statement.

This sounds like a temporary value doesn't live long enough. You did a great job

PR #268, opened 2021-10-11
Prettier docs

marioortizmanero:
## Description

We currently generate the docs for `docs.rs` with the reqwest client. Unfortunately, since that uses async-trait, the function signatures are horrifying:

image

So this just enables `client-ureq` for the docs:

image

Much better!

## Dependencies 

None

## Type of change

## How Has This Been Tested?

`cargo doc --no-default-features --features=cli,env-file,client-ureq --open`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,8 +78,11 @@ __async = ["futures", "async-stream", "async-trait"]
 __sync = ["maybe-async/is_sync"]
 
 [package.metadata.docs.rs]
-# Documenting the CLI methods, and working links for `dotenv`
-features = ["cli", "env-file"]
+# When generating the docs, we also want to include the CLI methods, and working
+# links for `dotenv`. We generate them for ureq so that the function signatures
+# of the endpoints don't look gnarly (because of `async-trait`).
+features = ["cli", "env-file", "client-ureq"]
+default-features = false
 
 [[example]]
 name = "client_creds"


marioortizmanero:
I know this is horribly late and it's like the third time I've gone in circles, but I'm reconsidering the trait-based inheritance auth structure. I was writing up a blog article about the new version, and as I started describing its disadvantages again, I realized that it's not that much of a great solution. Here's an extract:

* Since the functionality is provided by the trait, and not the client, the user will always need to import the trait in order to use the endpoints. This can be fixed somewhat easily by [providing a prelude](https://stackoverflow.com/questions/36384840/what-is-the-prelude) with whatever traits are necessary, so that the user can just do `use rspotify::prelude::*` and forget about it.
* You can't really have private functions in the trait. For instance, we have a `endpoint_get` method that wraps `http.get` to include the authentication headers used for the endpoints. It's defined in the trait because it's useful for every client, but the end user doesn't really need access to it; it's just noise in the documentation. There are two possible workarounds, so it's not a big deal:
  * Declare the method in the trait with `#[doc(hidden)]` so that it doesn't appear in the documentation.
  * Use a private function outside the trait instead.
* You can't use `-> impl Trait` in trait methods as of Rust 1.55, which we unfortunately need. Specially if you're using an asynchronous client, because async trait methods are `-> impl Future` after all. We can erase these types with the {{< crate async-trait >}} crate for now, until this is fixed thanks to GATs being implemented, which isn't too far off.

Not sure what to think honestly, lol.

PR #271, opened 2021-10-14
Add READMEs

ramsayleung:
## Description

It is unexpectable that the reason of why CI flow of publishing failed is the lack of `README.md`, just adding `README.md` for all crates. Just an old Chinese proverb saying:

"good things never come easy" 

https://github.com/ramsayleung/rspotify/runs/3896242685?check_suite_focus=true

## Motivation and Context

Fix CI flow

### DIFF
--- a/rspotify-http/Cargo.toml
+++ b/rspotify-http/Cargo.toml
@@ -6,7 +6,6 @@ authors = [
 name = "rspotify-http"
 version = "0.11.0"
 license = "MIT"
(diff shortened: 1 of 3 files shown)


ramsayleung:
Great idea!

marioortizmanero:
I've removed the previous v0.11 attempt. Try releasing the tag again and let's hope that does the trick.

PR #274, opened 2021-10-14
Fix docs (again)

marioortizmanero:
Turns out that for whatever reason it's `no-default-features = true` instead of `default-features = false`: https://docs.rs/about/metadata


marioortizmanero:
I've released the version myself, let's see if this does the trick...

PR #293, opened 2021-12-11
Unbound Send marker if compile target is wasm32

### LINKED ISSUE #292: Support for wasm and Cloudflare Workers
I would like to use this library with Cloudflare Workers.

when I try to compile the following code 
```
cargo check --target=wasm32-unknown-unknown
```
then it occurs this error
```
error: future cannot be sent between threads safely
   --> rspotify-http\src\reqwest.rs:110:38
    |
110 |       ) -> Result<String, Self::Error> {
    |  ______________________________________^
111 | |         self.request(Method::GET, url, headers, |req| req.query(payload))
112 | |             .await
113 | |     }
    | |_____^ future created by async block is not `Send`
    |
    = help: within `impl Future`, the trait `Send` is not implemented for `*mut u8`
note: future is not `Send` as this value is used across an await
   --> rspotify-http\src\reqwest.rs:93:13
    |
89  |         let response = request.send().await?;
    |             -------- has type `Response` which is not `Send`
...
93  |             response.text().await.map_err(Into::into)
    |             ^^^^^^^^^^^^^^^^^^^^^ await occurs here, with `response` maybe used later
...
97  |     }
    |     - `response` is later dropped here
    = note: required for the cast to the object type `dyn Future<Output = Result<std::string::String, ReqwestError>> + Send`
```

and when i use rspotify whth features=["client-ureq","ureq-rustls-tls"] on wrangler　project
I can compile, but I can't deploy with this error.
```
Error: Something went wrong with the request to Cloudflare...
Uncaught Error: LinkError: WebAssembly.Instance(): Import #21 module="env" function="GFp_ChaCha20_ctr32" error: function import requires a callable
  at line 0
 [API code: 10021]
```

I don't think there is anything that can't be done since reqwest also supports wasm.

How do you think?
Thaks.

tomocrafter:
## Description

Unbounds Send marker if compile target is wasm32 to fix #292

## Motivation and Context

Please also include relevant motivation and context. 

## Dependencies 

List any dependencies that are required for this change.

## Type of change

## How has this been tested?

This PR doesn't updated document. and I'm not used to write document in English proper so if anyone can, please create PR.

TBD

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,3 +91,37 @@ jobs:
         with:
           command: test
           args: -p rspotify -p rspotify-http -p rspotify-model -p rspotify-macros --no-default-features --features=${{ matrix.features }}
+          
+  wasm-test:
+    name: Test for wasm build
+    runs-on: ubuntu-latest
+    env:
+      RSPOTIFY_CLIENT_ID : e1dce60f1e274e20861ce5d96142a4d3
+      RSPOTIFY_CLIENT_SECRET : 0e4e03b9be8d465d87fc32857a4b5aa3
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Install stable toolchain for wasm
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          target: wasm32-unknown-unknown
+          profile: minimal  
+          override: true
+
+      - name: Run cargo check
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --target wasm32-unknown-unknown
+
+      - name: Install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+
+      - name: Run wasm-pack test firefox
+        run: wasm-pack test --headless --firefox
+
+      - name: Run wasm-pack test chrome
+        run: wasm-pack test --headless --chrome
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ rspotify-http = { path = "rspotify-http", version = "0.11.3", default-features =
 async-stream = { version = "0.3.2", optional = true }
 async-trait = { version = "0.1.51", optional = true }
 base64 = "0.13.0"
-chrono = { version = "0.4.19", features = ["serde"] }
+chrono = { version = "0.4.19", features = ["serde","wasmbind"] }
 dotenv = { version = "0.15.0", optional = true }
 futures = { version = "0.3.17", optional = true }
 log = "0.4.14"
@@ -50,6 +50,7 @@ webbrowser = { version = "0.5.5", optional = true }
 env_logger = { version = "0.9.0", default-features = false }
 tokio = { version = "1.11.0", features = ["rt-multi-thread", "macros"] }
 futures-util = "0.3.17"
+wasm-bindgen-test = "0.3.0"
 
 [features]
 default = ["client-reqwest", "reqwest-default-tls"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,10 +34,9 @@ rspotify-http = { path = "rspotify-http", version = "0.11.3", default-features =
 async-stream = { version = "0.3.2", optional = true }
 async-trait = { version = "0.1.51", optional = true }
 base64 = "0.13.0"
-chrono = { version = "0.4.19", features = ["serde", "rustc-serialize"] }
+chrono = { version = "0.4.19", features = ["serde"] }
 dotenv = { version = "0.15.0", optional = true }
 futures = { version = "0.3.17", optional = true }
-getrandom = "0.2.3"
 log = "0.4.14"
 maybe-async = "0.2.6"
 serde = { version = "1.0.130", default-features = false }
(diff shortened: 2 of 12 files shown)


marioortizmanero:
Oh interesting

marioortizmanero:
Oh interesting, so you don't need to toggle `Send` for these: https://github.com/ramsayleung/rspotify/blob/master/src/clients/base.rs#L262?

marioortizmanero:
Also, are you familiar with how GitHub Actions work? Could you perhaps add `wasm32-unknown-unknown` to our tests? I can help you out otherwise, but another day.

marioortizmanero:
And finally, it'd be nice if you could add some `wasm32-unknown-unknown` specific tests if possible. You can write them similarly to however you're using `rspotify` yourself for your project.

ramsayleung:
As Mario mentioning above, you could take the original cross-compile CI [configuration](https://github.com/ramsayleung/rspotify/pull/213/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f)(which has been removed) as an example if you have no idea how does our CI work

shiguma127:
Created a test.
This is my first time using HithubActions CI, so there are some things that are not good.
If you have any other ideas or good ways to do it, please let me know.
Thank you.

ramsayleung:
You could convert this PR to the stage of `ready for review`, and then the GitHub Action CI will run. Later, You could check if it works.

marioortizmanero:
Oh, nice. I was going to ask you to use `matrix` as we did before, but it seems the tests are ran in a very different way than with wasm, so I think it's ok in this case to add a new flow from scratch. I've approved the CI run to see if it passes, but I still wouldn't merge this PR until we review it in detail.

rspotify-model/Cargo.toml
```diff
@@ -14,7 +14,7 @@ edition = "2018"
 readme = "../README.md"
 
 [dependencies]
-chrono = { version = "0.4.19", features = ["serde", "rustc-serialize"] }
+chrono = { version = "0.4.19", features = ["serde","wasmbind"] }
```
marioortizmanero, line 17: AFAIK this enables `wasmbind` *always*, even when it's not compiling for `wasm32`. Perhaps it'd be better to use one of these `cfg` statements to only add the feature for `wasm32`?

tests/test_with_credential.rs
```diff
@@ -22,14 +28,33 @@ pub async fn creds_client() -> ClientCredsSpotify {
     spotify.request_token().await.unwrap();
     spotify
 }
+#[cfg(target_arch = "wasm32")]
```
marioortizmanero, line 31: Why not use the regular implementation? It seems like this does the same thing.

tests/test_with_credential.rs
```diff
@@ -22,14 +28,33 @@ pub async fn creds_client() -> ClientCredsSpotify {
...
+    };
+    let mut spotify = ClientCredsSpotify::new(creds);
+    spotify.request_token().await.unwrap();
+    spotify
+}
 
-#[maybe_async::test(feature = "__sync", async(feature = "__async", tokio::test))]
+#[cfg_attr(
+    not(target_arch = "wasm32"),
+    maybe_async::test(feature = "__sync", async(feature = "__async", tokio::test))
```
marioortizmanero, line 45: Can you try with a simpler approach like this one?

```rust
#[maybe_async::test(or(target_arch = "wasm32", feature = "__sync"), async(feature = "__async", tokio::test))]
```

Cargo.toml
```diff
@@ -34,10 +34,9 @@ rspotify-http = { path = "rspotify-http", version = "0.11.3", default-features =
 async-stream = { version = "0.3.2", optional = true }
 async-trait = { version = "0.1.51", optional = true }
 base64 = "0.13.0"
-chrono = { version = "0.4.19", features = ["serde", "rustc-serialize"] }
+chrono = { version = "0.4.19", features = ["serde","wasmbind"] }
```
marioortizmanero, line 37: Same as the other comment, this should maybe use `cfg`

.github/workflows/ci.yml
```diff
@@ -91,3 +91,37 @@ jobs:
         with:
           command: test
           args: -p rspotify -p rspotify-http -p rspotify-model -p rspotify-macros --no-default-features --features=${{ matrix.features }}
+          
+  wasm-test:
+    name: Test for wasm build
+    runs-on: ubuntu-latest
+    env:
+      RSPOTIFY_CLIENT_ID : e1dce60f1e274e20861ce5d96142a4d3
+      RSPOTIFY_CLIENT_SECRET : 0e4e03b9be8d465d87fc32857a4b5aa3
```
ramsayleung, line 100: Is it necessary to use these environment variables to run the test? Moreover, it's better to store these values as [Action secrets](https://docs.github.com/en/actions/security-guides/encrypted-secrets) instead of defining in configuration file explicitly(it's dangerous to leak this information to the whole world)

tests/test_with_credential.rs
shiguma127: Browser that runs wasm cannot fetch environment variables, we need to determine the value of the Credentials at compile time.
For this purpose, we have written another function using the ```env!``` macro.
Is this not a good idea?

PR #294, opened 2021-12-25
Update to rust edition 2021

### LINKED ISSUE #275: Use Rust edition 2021
Rust 1.56 has been released, and with it, the 2021 edition. There's lots of cool changes. The problem is that if we started using it right now we'd force all the users of Rspotify to use rustc 1.56, so we should leave it for the future.

Reference: https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html

* We should probably add the `rust-version` field to the `Cargo.toml`s
* We should remove `resolver = "2"` from the `Cargo.toml`s, since that's default now
* Try removing the `build_map` and `build_json` macros now that `HashMap::from([(a,b), (c,d)])` is possible (only in case it looks great).
* Any other things that come up. We should run `cargo fix --edition` just in case.

FWIW, Rspotify seems to compile just fine with v1.56, so that's not a problem.

ramsayleung:
## Description

Update to  Rust edition 202

## Motivation and Context

Trying to close #275, by the way, as for this:

> Try removing the build_map and build_json macros now that HashMap::from([(a,b), (c,d)]) is possible (only in case it looks great).

I think it's necessary to keep `build_json` and `build_map` macros, since the current `optional` semantics can't  be replaced with the built-in `HashMap::from` method:

```rust
        let params = build_map! {
            optional "album_type": album_type.map(|x| x.as_ref()),
            optional "market": market.map(|x| x.as_ref()),
            optional "limit": limit.as_deref(),
            optional "offset": offset.as_deref(),
        };
```

## Dependencies 

None

## Type of change

## How has this been tested?

All CI checks should pass

I think it's not necessary to add this to CHANGELOG

Last but not least, Merry Christmas :-)


marioortizmanero:
You forgot to edit the edition for the rest of the crates. We should probably add `rust-version = "1.56"` to the `Cargo.toml` while we're at it. And remove `resolver = "2"`, since that's default now.

> I think it's necessary to keep build_json and build_map macros, since the current optional semantics can't be replaced with the built-in HashMap::from method:

Oh, you're right, I completely forgot about that. Yeah, I'd rather keep the macros in that case.

Anyway, it might be better to wait a bit more, maybe until 1.58 is released?

Merry Christmas, Ramsay!

ramsayleung:
> You forgot to edit the edition for the rest of the crates. We should probably add rust-version = "1.56" to the Cargo.toml while we're at it. And remove resolver = "2", since that's default now.

Good point!

> Anyway, it might be better to wait a bit more, maybe until 1.58 is released?

Sure, we could wait for 1.58

marioortizmanero:
It's time to merge this for the next version :)

marioortizmanero (approved):

PR #298, opened 2022-02-17
First draft for write pagination

### LINKED ISSUE #296: Automatically do paginated post requests

After gathering some information from the Spotify API, I tried to add around 300 tracks to a single playlist. I knew that Spotify restricts the number of items that can be passed to an endpoint. For the [Add Items to Playlist](https://developer.spotify.com/documentation/web-api/reference/#/operations/add-tracks-to-playlist) endpoint, Spotify imposes a limit of a 100 tracks in a single request. The API does not reflect this restriction in any way.

While it would be possible to add some type information to functions like [playlist_add_items](https://github.com/ramsayleung/rspotify/blob/52e35c8e4136e4895b0514766a44a2121bc498ba/src/clients/oauth.rs#L296), I would rather prefer that all method do chunking automatically. If the given number of items surpasses the limit of the API, a second request will be sent.

It is quite easy to implement the chunking on my own. However, it is dull and the returned responses with a status code of 400 might surprise some users.

**Additional context**

I am quite new to Rust and, given you agree to this proposal, I would love to create a PR for this.

nick-lehmann:
## Description

Add a method to paginate write requests to the Spotify API. Fixes #296.

This is currently a draft. I'm still quite new to Rust and struggle with `async` Rust code. At the moment, the changes do not work and I'm looking for some feedback by someone more experienced than me.

## Motivation and Context

Spotify imposes a limit on items not only for reading endpoints but also for endpoints that create data. Currently, the api client will simply fail in case the given items exceed the limit. This PR adds pagination for methods like `playlist_add_items`.

## Dependencies 

None

## Type of change

## How has this been tested?

Adapted integration tests to check that data is successfully created even if the items exceeds the limits of the Spotify API. 

Not yet. Will do that as soon as the feature is complete.


nick-lehmann:
@marioortizmanero @ramsayleung it would be very kind of you if you would have a quick look at the approach I have taken. At the moment, the code does not work because I pass a closure to my `write_paginate` function. The closure gets the current chunk of items, transform them into the payload and executes the `endpoint_post` method. 

However, the `self` reference to the oauth client ([here](https://github.com/nick-lehmann/rspotify/blob/master/src/clients/oauth.rs#L299)) does not live long enough to do this. I was unable to fix this issue. May you please give me a hint where I got things wrong and how I could improve on my design? I hope I did formulate my problem adequatly 😅

marioortizmanero:
It would probably be easier if you started with the synchronous version, and then moved on to the asynchronous one. The `self` borrow issue can be solved by using generics for the `Future` instead:

```rust
pub async fn write_paginate<'a, Item, Writer: 'a, Fut>(
    writer: Writer,
    items: Vec<Item>,
    chunk_size: u32,
) -> String
where
    Writer: Fn(&[Item], u32) -> Fut,
    Fut: futures::Future<Output = Result<String, ClientError>>,
    Item: Display + Debug,
```

The problem then becomes that you're creating `params` inside the function, and it's borrowed by the future you're returning. `params` would  drop at the end of the closure, but you haven't awaited the future. Option 1 would be to create params separately. Option 2 is to await the future in the closure, but that would require async closures, which are  not stable yet. Option 3 would be to make `write_paginate` a macro instead. But I think it's a design issue, maybe it can be done with a different approach, also knowing these other things:

* Don't unwrap, propagate the error
* The returned values should be combined. Returning the last one only doesn't make sense for e.g. `current_user_saved_albums_contains`.
* Maybe different naming? Not sure what `write_paginate` refers to.

PR #325, opened 2022-06-15
Remove undocumented modules containing one item

SabrinaJewson:
## Description

This removes the modules `auth_code`, `auth_code_pkce`, `client_creds`, `clients::base` and `clients::oauth` from the public API surface.

## Motivation and Context

These modules are unnecessary because they only export a single item which has the same name as the module itself anyway. Documentation is easier to navigate if they’re not present.

## Dependencies 

None.

## Type of change

## How has this been tested?

- [ ] Will add a changelog entry once I have a PR number.

### DIFF
--- a/src/clients/mod.rs
+++ b/src/clients/mod.rs
@@ -1,6 +1,6 @@
-pub mod base;
+mod base;
 pub mod mutex;
-pub mod oauth;
+mod oauth;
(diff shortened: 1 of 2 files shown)


marioortizmanero:
Awesome! That should clean up the docs a bit.

Thanks for the PR!

PR #326, opened 2022-06-15
Move `clients::mutex` to `sync` module

SabrinaJewson:
## Description

Rename `clients::mutex` to `sync` and add doc comments to it.

## Motivation and Context

`clients::mutex` shouldn’t really be inside `clients` IMO because it has nothing inherently to do with the differerent clients. I was considering removing the module entirely, but I think it’s nice to keep it contained. `sync` is a better name than `mutex` since it’s more general.

## Dependencies 

None.

## Type of change

## How has this been tested?

- [x] Test A: `cargo test --no-default-features --features client-reqwest,reqwest-default-tls,env-file`
- [x] Test B: `cargo test --no-default-features --features client-ureq,ureq-rustls-tls,env-file`

I will add a changelog entry.


marioortizmanero:
Looks good to me! The only nitpick I have is that maybe `sync::sync` should be renamed to `sync::blocking`? It seems less confusing.

SabrinaJewson:
Good point, I’ve done that. I do run Clippy locally but it didn’t complain about it for some reason…how strange.

marioortizmanero:
Yeah, since it's a workspace you have to use `cargo clippy --all` or something like that. I think. I'll merge after you do a `cargo fmt`, it seems like the imports aren't sorted anymore :P

marioortizmanero (approved):

PR #327, opened 2022-06-15
Remove `ModelError` in favour of strongly-typed errors

SabrinaJewson:
## Description

* Removed `ModelError`
* Added `ReadTokenCacheError` and `WriteTokenCacheError`
* To support the above two types, added `ReadFileError` and `WriteFileError`
* Replaced `ClientError::Model` with the more explicit and useful `ClientError::UpdateTokenCache`
* Also removed `ClientError::CacheFile`, which seems to have been left in by accident

## Motivation and Context

Strongly-typed errors are better than weakly typed errors because:
* They make it easier for the user to see exactly what can go wrong through just types
* They give better error messages

For example, printing the backtrace of one of the new error types might give this:

```
Error: failed to read token from cache

Caused by:
0: failed to read file token_cache.json
1: No such file or directory (os error 2)
```

Which is extremely helpful to the user.

## Dependencies 

None.

## Type of change

## How has this been tested?


src/client_creds.rs
```diff
@@ -93,7 +94,7 @@ impl ClientCredsSpotify {
     /// * The read token is expired
     /// * The cached token is disabled in the config
     #[maybe_async]
-    pub async fn read_token_cache(&self) -> ClientResult<Option<Token>> {
+    pub async fn read_token_cache(&self) -> Result<Option<Token>, ReadTokenCacheError> {
```
marioortizmanero, line 97: This change can be avoided by making this error part of `ClientError`

marioortizmanero, line 97: This change can be avoided by making this error part of `ClientError`

src/clients/oauth.rs
```diff
@@ -56,7 +56,10 @@ pub trait OAuthClient: BaseClient {
     /// the application re-authenticate.
     ///
     /// [`ClientCredsSpotify::read_token_cache`]: crate::client_creds::ClientCredsSpotify::read_token_cache
-    async fn read_token_cache(&self, allow_expired: bool) -> ClientResult<Option<Token>> {
+    async fn read_token_cache(
+        &self,
+        allow_expired: bool,
+    ) -> Result<Option<Token>, ReadTokenCacheError> {
```
marioortizmanero, line 62: This change can be avoided by making this error part of `ClientError`

marioortizmanero (commented):
I love the idea, thanks for submitting a PR.

* Why not use `#[derive(Error)]` instead of implementing them manually, though? We already have the dependency anyway.
* Shouldn't `ReadTokenCacheError` and etc be a variant in `ClientError`?

src/client_creds.rs
SabrinaJewson: But then all code that deals in `ClientError`s would now have one more variant to think about even though it can’t actually be created from most code, and similarly any code that calls this function now has to deal with every variant of `ClientError` even though only one actually can be created.

I suppose it would technically fit `ClientError`’s stated purpose to use it here, but the API can be made more strongly typed by not using it I think.

rspotify-model/src/auth.rs
```diff
@@ -95,6 +93,109 @@ impl Token {
...
+        inner,
+        path: Box::from(path),
+    })
+}
+
+/// An error reading a file.
+#[derive(Debug, Error)]
+#[error("failed to read file {}", path.display())]
+pub struct ReadFileError {
+    // Intentionally not exposed to allow future API evolution, e.g. moving this to a enum variants
```
marioortizmanero, line 128: Nitpick: there should be a comma after `e.g.`. I think this also breaks the 80-character rule.

src/client_creds.rs
marioortizmanero: Fair enough. My reasoning was that if `ReadTokenCacheError` isn't in `ClientError`, then if someone used to return a `ClientError` will now have to define their own custom error type. But I guess that's not really a huge issue, as most libraries will have their own error anyway.

rspotify-model/src/auth.rs
```diff
@@ -95,6 +93,109 @@ impl Token {
...
+    /// There was an error reading the cache file into memory.
+    Reading(#[source] ReadFileError),
+
+    /// There was an error deserializing the contents of the cache file.
+    Deserializing(#[source] serde_json::Error),
+}
+
+/// An error writing a [`Token`] to its cache.
+#[derive(Debug, Error)]
+#[non_exhaustive]
```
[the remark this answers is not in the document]
SabrinaJewson: Could’ve been a tuple struct I suppose — should I change it?

rspotify-model/src/auth.rs
SabrinaJewson: Oh huh, that’s an American thing.

marioortizmanero: Is it? I might be wrong, though. I've looked into it when I was writing papers or similars, and it makes sense to me, since you could replace `e.g.` with `for example`, which is usually followed by a comma (https://english.stackexchange.com/questions/16172/should-i-always-use-a-comma-after-e-g-or-i-e).

rspotify-model/src/auth.rs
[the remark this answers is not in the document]
marioortizmanero: I'd say we don't really need to future-proof anything here? Not sure what else we would need to add into the error.

rspotify-model/src/auth.rs
marioortizmanero: Ah yeah, seems like it's more British:

> So, in short, I am happy to say that, as a Brit, I am not wrong in my refusal to add the comma, but then again, neither are the subsequent editors of the text. It just depends upon which flavour of English one uses natively. ;-)

Not like it matters for this library anyway. I'm neither British nor American, so my own English is just a weird mix of both.

PR #329, opened 2022-06-15
Replace `build_json!` with a JSON builder API

SabrinaJewson:
## Description

Instead of:

```rs
build_json! {
    "a": b,
    optional "c": d,
}
```

we now use:

```rs
json_builder()
    .field("a", b)
    .optional_field("c", d)
    .build()
```

I would’ve used an array-based API like in #328 but that doesn’t really work well here because the field value types can be any type rather than just strings.

## Dependencies 

None.

## How has this been tested?

### DIFF
--- a/rspotify-macros/src/lib.rs
+++ b/rspotify-macros/src/lib.rs
@@ -39,57 +39,6 @@ macro_rules! count_items {
     ($($item:expr),*) => {<[()]>::len(&[$($crate::replace_expr!($item ())),*])};
 }
 
-/// This macro and [`build_json`] help make the endpoints as concise as possible
-/// and boilerplate-free, which is specially important when initializing the
-/// parameters of the query. In the case of `build_map` this will construct a
-/// `HashMap<&str, &str>`, and `build_json` will initialize a
-/// `HashMap<String, serde_json::Value>`.
-///
-/// The syntax is the following:
-///
-///   [optional] "key": value
-///
-/// For an example, refer to the `test::test_build_map` function in this module,
-/// or the real usages in Rspotify's client.
-///
-/// The `key` and `value` parameters are what's to be inserted in the HashMap.
-/// If `optional` is used, the value will only be inserted if it's a
-/// `Some(...)`.
-#[doc(hidden)]
-#[macro_export]
-macro_rules! internal_build_map {
-    (/* required */, $map:ident, $key:expr, $val:expr) => {
-        $map.insert($key, $val);
-    };
-    (optional, $map:ident, $key:expr, $val:expr) => {
-        if let Some(val) = $val {
-            $map.insert($key, val);
-        }
-    };
-}
-#[doc(hidden)]
-#[macro_export]
-macro_rules! build_map {
-    (
-        $(
-            $( $kind:ident )? $key:literal : $val:expr
-        ),+ $(,)?
-    ) => {{
-        let mut params = ::std::collections::HashMap::<&str, &str>::with_capacity(
-            $crate::count_items!($( $key ),*)
-        );
-        $(
-            $crate::internal_build_map!(
-                $( $kind )?,
-                params,
-                $key,
-                $val
-            );
-        )+
-        params
-    }};
-}
-
 /// Refer to the [`build_map`] documentation; this is the same but for JSON
 /// maps.
 #[doc(hidden)]
@@ -129,9 +78,8 @@ macro_rules! build_json {
 
 #[cfg(test)]
 mod test {
-    use crate::{build_json, build_map, scopes};
+    use crate::{build_json, scopes};
     use serde_json::{json, Map, Value};
-    use std::collections::HashMap;
 
     #[test]
     fn test_hashset() {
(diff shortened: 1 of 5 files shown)


marioortizmanero (commented):
Oh, nice! A builder pattern is a great solution, too. Thoughts: 

* What about renaming `field` to `required` and `field_optional` to just `optional`? Easier to read, IMO.
* Why not use `JsonBuilder::new` instead of `json_builder`? That looks more like a regular builder pattern.
* What if we went for a similar approach for the map version? Just for "consistency", and we could possibly use some magic that avoids all these `.map(|x| x.as_ref())` and `.map(|x| x.as_deref())`. Not quite sure how that could be done, though, just an idea. Might not be worth it in the end, but we could explore it.

SabrinaJewson:
Well, it would actually be possible to not affect performance. You just need to leverage type inference a bit:

```rs
pub(crate) struct JsonBuilder<Len> {
    map: serde_json::Map<String, serde_json::Value>,
    len: PhantomData<fn() -> Len>,
}

impl<Len: ToUsize> JsonBuilder<Len> {
    pub(crate) fn new() -> Self {
        Self {
            map: serde_json::Map::with_capacity(Len::to_usize()),
            len: PhantomData,
        }
    }
}

impl<Len: ToUsize> JsonBuilder<S<Len>> {
    pub(crate) fn required(mut self, name: &str, value: impl Serialize) -> JsonBuilder<Len> {
        self.map
            .insert(name.to_owned(), serde_json::to_value(value).unwrap());
        JsonBuilder {
            map: self.map,
            len: PhantomData,
        }
    }

    pub(crate) fn optional(self, name: &str, value: Option<impl Serialize>) -> JsonBuilder<Len> {
        if let Some(value) = value {
            self.required(name, value)
        } else {
            JsonBuilder {
                map: self.map,
                len: PhantomData,
            }
        }
    }
}

impl JsonBuilder<Z> {
    pub(crate) fn build(self) -> serde_json::Value {
        serde_json::Value::Object(self.map)
    }
}

pub(crate) struct Z;

pub(crate) struct S<T>(T);

pub(crate) trait ToUsize {
    fn to_usize() -> usize;
}

impl ToUsize for Z {
    fn to_usize() -> usize {
        0
    }
}

impl<T: ToUsize> ToUsize for S<T> {
    fn to_usize() -> usize {
        T::to_usize() + 1
    }
}
```

marioortizmanero:
Hah, awesome, very haskell-y. I was thinking of just using const generics but it's even better like that. Is the `PhantomData<fn() -> Len>` to make the type covariant over `Len`? But wouldn't `PhantomData<Len>` do the same?

I would say this is much better than the macro. Perhaps not as understandable if you aren't used to const generics, but macros are pretty weird, too. As long as it's nicely commented I would say it's great. I would also rename `Z` to `Zero` and `S` to `Natural`, since I've only seen that naming in Haskell (I assume you were inspired by it? Or just pure Rust?). And maybe `ToUsize` could be `IncrUsize` to emphasize on how it increments the constant whenever it's called?

Props for such a neat design!!

SabrinaJewson:
> Is the `PhantomData<fn() -> Len>` to make the type covariant over `Len`? But wouldn't `PhantomData<Len>` do the same?

I use `PhantomData<fn() -> Len>` out of principle, because it is usually the more desired type. To me, `PhantomData<Len>` sticks out in code as “something funny’s going on here” because usually you don’t want to inherit `Send` and `Sync` bounds from the inner type. But I could change it easily.

> I assume you were inspired by it?

I was inspired by [Peano arithmetic](https://en.wikipedia.org/wiki/Peano_axioms), in which there is a constant `Z` and a function `S`. But yeah, probably better to use more descriptive names. I’ll rename to `Zero` and `Successor`.

> And maybe `ToUsize` could be `IncrUsize` to emphasize on how it increments the constant whenever it's called?

I think `ToUsize` is a good name, because conceptually a type like `S<S<Z>>` represents `2`, so calling `<S<S<Z>>>::to_usize()` should yield `2`.

---

Now I’ve actually implemented it, and I named `ToUsize` `Natural` as well as made the value an associated constant.

marioortizmanero:
> I use PhantomData<fn() -> Len> out of principle, because it is usually the more desired type. To me, PhantomData<Len> sticks out in code as “something funny’s going on here” because usually you don’t want to inherit Send and Sync bounds from the inner type. But I could change it easily.

No worries. It especially makes sense in this case; `PhantomData<Len>` would probably be better if we had a pointer to `Len` or something like that.

I would say that a comment explaining the `fn() -> Len` would be nice, but it might be that I just haven't seen that myself before personally.

> I was inspired by Peano arithmetic, in which there is a constant Z and a function S. But yeah, probably better to use more descriptive names. I’ll rename to Zero and Successor.
> I think ToUsize is a good name, because conceptually a type like S<S<Z>> represents 2, so calling <S<S<Z>>>::to_usize() should yield 2.

Cool, I hadn't seen those in Rust yet. I think the naming is spot-on now. I also prefer it with associated constants.

marioortizmanero (approved):
Really nicely commented! It seems like this branch is over `map-builder`, so we would need `MapBuilder` implemented before merging this PR.

SabrinaJewson:
Sorry about forgetting about this one!

marioortizmanero:
No worries, thank you very much for the contribution! I'm merging this.

PR #330, opened 2022-06-15
Avoid accepting references to `Copy` types

SabrinaJewson:
## Description

- `search` accepts `Option<IncludeExternal>` instead of `Option<&IncludeExternal>`
- `featured_playlists` accepts `Option<chrono::DateTime<chrono::Utc>>` instead of `Option<&chrono::DateTime<chrono::Utc>>`
- `current_user_top_artists[_manual]` and `current_user_top_tracks[_manual]` accept `Option<TimeRange>` instead of `Option<&TimeRange>`

I didn’t do the same for `Market` parameters because they’re not `Copy` (although they could be) — I’ll leave it to a separate PR.

## Motivation and Context

Accepting a reference to a `Copy` type is usually counterproductive, since you can just take it by value. For most types this makes the size of the parameters smaller, for the `Option<chrono::DateTime<chrono::Utc>>` it increases it from 8 bytes to 16 bytes but I think that’s acceptable.

## Dependencies 

None.

## Type of change

## How has this been tested?

### DIFF
--- a/src/clients/base.rs
+++ b/src/clients/base.rs
@@ -436,7 +436,7 @@ where
         q: &str,
         _type: &SearchType,
         market: Option<&Market>,
-        include_external: Option<&IncludeExternal>,
+        include_external: Option<IncludeExternal>,
         limit: Option<u32>,
         offset: Option<u32>,
     ) -> ClientResult<SearchResult> {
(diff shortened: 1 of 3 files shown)


marioortizmanero:
Also, @ramsayleung, I'm merging the small PRs that are just nice to have and don't require discussion (like this one). But I'll leave the rest open to wait for your approval as well, in case you come up with a better solution or with any ideas.

ramsayleung:
> I'm merging the small PRs that are just nice to have and don't require discussion (like this one). But I'll leave the rest open to wait for your approval as well, in case you come up with a better solution or with any ideas.

Everything looks good to me too :)

PR #331, opened 2022-06-15
Implement `Into<&'static str>` for all enums

SabrinaJewson:
## Description

Derives both `AsRef<str>` and `Into<&'static str>` (via `From<{the enum}> for &'static str`) for all the enums.

Note: there’s a `#[derive]` in here that `rustfmt` explodes on to 13 lines. I don’t really like it but I also can’t separate the derives into two lines because rustfmt just merges them together again 😄

Also makes `Market` `Copy`, just so I can avoid code duplication between `AsRef<str>` and `From<Market> for &'static str`.

## Motivation and Context

This makes the enums useful in more places, since the lifetime of the string is not actually tied to the lifetime of the enum.

## Dependencies 

None.

## Type of change

## How has this been tested?

- [x] Test A: `cargo test -p rspotify-model`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+- ([#331](https://github.com/ramsayleung/rspotify/pull/331)) All enums now implement `Into<&'static str>` as well as `AsRef<str>`
+- ([#331](https://github.com/ramsayleung/rspotify/pull/331)) `Market` is now `Copy`
+
 **Breaking changes**:
 - ([#325](https://github.com/ramsayleung/rspotify/pull/325)) The `auth_code`, `auth_code_pkce`, `client_creds`, `clients::base` and `clients::oauth` modules have been removed from the public API; you should access the same types from their parent modules instead
 - ([#326](https://github.com/ramsayleung/rspotify/pull/326)) The `rspotify::clients::mutex` module has been renamed to `rspotify::sync`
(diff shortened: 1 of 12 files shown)


SabrinaJewson:
Note that this would allow lots of the endpoint functions to use `x.map(|x| x.into())` instead of `x.as_ref().map(|x| x.as_ref())` so that’s nice. And making `Market` `Copy` also opens up the door for changing all those `Option<&Market>` parameters to take `Option<Market>` instead; but I haven’t done that yet because the merge conflicts with #328 would be a pain.

marioortizmanero:
Yeah, `IntoStaticStr` seems better overall. Is there a point in also having `AsRefStr` now?

SabrinaJewson:
I suppose not. It might be useful occasionally — your call really.

marioortizmanero:
I would personally remove it, but let's wait for @ramsayleung to have a third opinion.

ramsayleung:
If we have `IntoStaticStr`, I think there is unnecessary to keep `AsRefStr`, as @SabrinaJewson points out:

> since the lifetime of the string is not actually tied to the lifetime of the enum.

SabrinaJewson:
I’ve gone ahead and removed those implementations, also replacing `&Market` types with `Market` to prevent `(*x).into()`s everywhere (this is possible now since `Market: Copy`).

PR #332, opened 2022-06-21
Fix: Typo of explicit restriction in `RestrictionReason`

hrkfdn:
## Description

Looks like there is a typo in the enum values. Not sure if this would considered a breaking change since the enum options are changing.

## Motivation and Context

https://github.com/hrkfdn/ncspot/issues/844

## Type of change

## How has this been tested?

See link above. I didn't explicitly test it, but it's surely a typo :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Unreleased
 
+**Bugfixes**:
+- ([#332](https://github.com/ramsayleung/rspotify/pull/332)) Fix typo in `RestrictionReason` enum values
+
 **Breaking changes**:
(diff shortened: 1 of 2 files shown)


hrkfdn:
Thanks @marioortizmanero, just out of curiosity: is a release planned anytime soon?

marioortizmanero:
I was hoping to merge #331, #329, #327 and #305 as a new 0.12 release, if that's fine with you. They should be done already and only need feedback before merging.

marioortizmanero:
Another option is to merge all of the PRs except for #305, which is the truly breaking & bigger change, and release a 0.11.6 release in the meanwhile. But it depends on how fast we can merge the other ones and etc.

hrkfdn:
Sooo @marioortizmanero, any plans for a bugfix release? :sweat_smile:

hrkfdn:
Thanks @ramsayleung ❤️

PR #337, opened 2022-07-13
Fix double referenced types in paginated endpoints

marioortizmanero:
## Description

This removes the potential double reference to Id types (&'a Id<'_>)
which was previously required due to limitations in the async variant of
the `paginate` function.
Due to limitations in HRTBs (see
https://kevincox.ca/2022/04/16/rust-generic-closure-lifetimes/) it is
not possible, to write something like `Req: for<'a> Fn(&'a) ->
('a + Future<...>)`. The sync version remains almost unchanged.

## Motivation and Context

See #305

## Dependencies 

None

## Type of change

## How has this been tested?

CI passes

No need, it's an internal change

### DIFF
--- a/src/clients/pagination/iter.rs
+++ b/src/clients/pagination/iter.rs
@@ -5,6 +5,17 @@ use crate::{model::Page, ClientError, ClientResult};
 /// Alias for `Iterator<Item = T>`, since sync mode is enabled.
 pub type Paginator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
 
+pub fn paginate_with_ctx<'a, Ctx: 'a, T: 'a, Request: 'a>(
+    ctx: Ctx,
+    req: Request,
+    page_size: u32,
+) -> Paginator<'a, ClientResult<T>>
+where
+    Request: Fn(&Ctx, u32, u32) -> ClientResult<Page<T>>,
+{
+    paginate(move |limit, offset| req(&ctx, limit, offset), page_size)
+}
+
 /// This is used to handle paginated requests automatically.
 pub fn paginate<'a, T: 'a, Request: 'a>(
     req: Request,
--- a/src/clients/pagination/stream.rs
+++ b/src/clients/pagination/stream.rs
@@ -9,7 +9,34 @@ use futures::{future::Future, stream::Stream};
 /// Alias for `futures::stream::Stream<Item = T>`, since async mode is enabled.
 pub type Paginator<'a, T> = Pin<Box<dyn Stream<Item = T> + 'a>>;
 
+pub type RequestFuture<'a, T> = Pin<Box<dyn 'a + Future<Output = ClientResult<Page<T>>>>>;
+
 /// This is used to handle paginated requests automatically.
+pub fn paginate_with_ctx<'a, Ctx: 'a, T, Request>(
+    ctx: Ctx,
+    req: Request,
+    page_size: u32,
+) -> Paginator<'a, ClientResult<T>>
+where
+    T: 'a + Unpin,
+    Request: 'a + for<'ctx> Fn(&'ctx Ctx, u32, u32) -> RequestFuture<'ctx, T>,
+{
+    use async_stream::stream;
+    let mut offset = 0;
+    Box::pin(stream! {
+        loop {
+            let page = req(&ctx, page_size, offset).await?;
+            offset += page.items.len() as u32;
+            for item in page.items {
+                yield Ok(item);
+            }
+            if page.next.is_none() {
+                break;
+            }
+        }
+    })
+}
+
 pub fn paginate<'a, T, Fut, Request>(req: Request, page_size: u32) -> Paginator<'a, ClientResult<T>>
 where
     T: 'a + Unpin,
--- a/rspotify-model/src/album.rs
+++ b/rspotify-model/src/album.rs
@@ -33,7 +33,7 @@ pub struct SimplifiedAlbum {
 }
 
 /// Full Album Object
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct FullAlbum {
     pub artists: Vec<SimplifiedArtist>,
     pub album_type: AlbumType,
@@ -67,7 +67,7 @@ pub struct PageSimplifiedAlbums {
 }
 
 /// Saved Album object
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct SavedAlbum {
     pub added_at: DateTime<Utc>,
     pub album: FullAlbum,
--- a/rspotify-model/src/lib.rs
+++ b/rspotify-model/src/lib.rs
@@ -7,7 +7,7 @@ pub mod audio;
 pub mod auth;
 pub mod category;
 pub mod context;
-pub(in crate) mod custom_serde;
+pub(crate) mod custom_serde;
 pub mod device;
 pub mod enums;
 pub mod error;
--- a/rspotify-model/src/playlist.rs
+++ b/rspotify-model/src/playlist.rs
@@ -36,7 +36,7 @@ pub struct SimplifiedPlaylist {
 }
 
 /// Full playlist object
-#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct FullPlaylist {
     pub collaborative: bool,
     pub description: Option<String>,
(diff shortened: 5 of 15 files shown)


marioortizmanero:
Whoops, I think that by creating the PR myself @eladyn, the original author, can't commit themselves. Can you confirm that? Otherwise, you can create it yourself to iterate over it.

eladyn:
Unfortunately, I'm currently not at home and won't be able to get back to this for a week or so. If you'd like to pick this up and fix the tests for example in the meantime, feel free to do so. I'll be glad to assist and explain things as good as possible. :)

eladyn:
So I finally got around to fixing the tests. Let me know if you need to know anything about that implementation. It might be a bit confusing at first, but I didn't find any better way to do it.

marioortizmanero:
Thanks! I didn't end up having enough time to fix it myself. The [article you linked](https://kevincox.ca/2022/04/16/rust-generic-closure-lifetimes/) explains it quite well, though.

I wonder if this will be fixed in the future with something like [chalk](https://github.com/rust-lang/chalk). I guess we need to be able to use `impl` within the `FnOnce` return type, but not sure if that's something `chalk` takes care of.

I'll wait for @ramsayleung to take a look, since this is a bit complicated and he should understand it as well.

ramsayleung:
This PR is a little complex, I need a little time to understand the context and solution :-)

marioortizmanero:
@ramsayleung, just wanted to let you know that I'll be out for vacation for a couple weeks, so I won't be able to work on the next version myself. However, only this PR and #327 are required, you can ignore #356 for now. I already reviewed the two required PRs, so if you don't need any additional feedback you can just release the version yourself. Otherwise, I'll be back in a couple weeks.

ramsayleung:
Sure, just enjoy your summer vacation :)

eladyn:
I just realized that by doing `cargo fmt` on Rust `1.63.0`, all those occurrences of `pub(in crate)` were changed to `pub(crate)`. Should I split that off in a separate PR or is that minor enough to stay in this one?

ramsayleung:
> Should I split that off in a separate PR or is that minor enough to stay in this one?

I think the modification is minor so that you could just stay in this one, you don't need a separate PR :)

src/clients/pagination/iter.rs
```diff
@@ -5,6 +5,17 @@ use crate::{model::Page, ClientError, ClientResult};
 /// Alias for `Iterator<Item = T>`, since sync mode is enabled.
 pub type Paginator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
 
+pub fn paginate_with_ctx<'a, Ctx: 'a, T: 'a, Request: 'a>(
+    ctx: Ctx,
+    req: Request,
+    page_size: u32,
+) -> Paginator<'a, ClientResult<T>>
```
ramsayleung, line 12: I am a little confused that what's `ctx` for?

src/clients/pagination/stream.rs
```diff
@@ -9,7 +9,34 @@ use futures::{future::Future, stream::Stream};
 /// Alias for `futures::stream::Stream<Item = T>`, since async mode is enabled.
 pub type Paginator<'a, T> = Pin<Box<dyn Stream<Item = T> + 'a>>;
 
+pub type RequestFuture<'a, T> = Pin<Box<dyn 'a + Future<Output = ClientResult<Page<T>>>>>;
```
ramsayleung, line 12: I have to say, It's a little headache to understand this function signature 😂

ramsayleung:
I have fixed the `clippy` and `fmt` error in this PR #361, perhaps we could merge #361 first.

src/clients/pagination/stream.rs
```diff
@@ -9,7 +9,34 @@ use futures::{future::Future, stream::Stream};
...
+
 /// This is used to handle paginated requests automatically.
+pub fn paginate_with_ctx<'a, Ctx: 'a, T, Request>(
+    ctx: Ctx,
+    req: Request,
+    page_size: u32,
+) -> Paginator<'a, ClientResult<T>>
+where
+    T: 'a + Unpin,
+    Request: 'a + for<'ctx> Fn(&'ctx Ctx, u32, u32) -> RequestFuture<'ctx, T>,
```
ramsayleung, line 22: I am not sure if I understand the design correctly, so I try to retell the story.

What the signature means is to allow `req` function to hold a reference to `Ctx` in the returned future with [HRTB](https://doc.rust-lang.org/nomicon/hrtb.html) 

So that the returned value of the `req` function is allowed to reference the argument(`Ctx`), we pass it.

All the thing we want to do is to specify that the lifetime of `Ctx` parameter must outlive the lifetime of `RequestFuture`?

src/clients/pagination/iter.rs
eladyn: In the sync case: not much. This is only to keep compatibility with the async version, which needs those. I'll try to explain why on the other review comments.

src/clients/pagination/stream.rs
eladyn: I see what you mean, coming up with it wasn't easy either. :grinning: In essence it's mostly the signature of a `Future` that returns a `ClientResult<Page<T>>` with the speciality that it is `Box`ed (and pinned) here.

src/clients/pagination/stream.rs
eladyn: That sounds correct. I'll add my version, so you can double check. :)

The problem we need to solve is: When we move anything into the closure (`req`), the returned future must not reference it in some way, since we can't ensure that the closure outlives the returned future.

The way we solve it: We pass the reference from outside and await the future nearly immediately afterwards. As such, the lifetime bounds are upheld and the result we get no longer references `ctx` so we can safely stream it to the consumer.

Due to the limitations in HRTBs, we can't formulate a bound at this time that works with a raw `impl Future<...>` so we need to `Box` it (which creates the monstrosity from your above review comment :sweat_smile:).

ramsayleung: I am a little confused about this:

> We pass the reference from outside and await the future nearly immediately afterwards. As such, the lifetime bounds are upheld

```rust
    fn user_playlists<'a>(
        &'a self,
        user_id: UserId<'a>,
    ) -> Paginator<'_, ClientResult<SimplifiedPlaylist>> {
        paginate_with_ctx(
            (self, user_id),
            move |(slf, user_id), limit, offset| {
                slf.user_playlists_manual(user_id.as_ref(), Some(limit), Some(offset))
            },
            self.get_config().pagination_chunks,
        )
    }
```

We pass the `(self, user_id)` as the `ctx`, but how does the lifetime bound? 

>  and the result we get no longer references ctx so we can safely stream it to the consumer.

According to the function signature: `for<'ctx> Fn(&'ctx Ctx, u32, u32)`, the closure references to `ctx`, do you mean the result `RequestFuture` doesn't reference to `ctx`?

eladyn: > I am a little confused about this:
> 
> > We pass the reference from outside and await the future nearly immediately afterwards. As such, the lifetime bounds are upheld
> 
> ...
> We pass the `(self, user_id)` as the `ctx`, but how does the lifetime bound?

What I meant by that is:

```rust
pub fn paginate_with_ctx<'a, Ctx: 'a, T, Request>(...) -> Paginator<'a, ClientResult<T>>
where
    T: 'a + Unpin,
    Request: 'a + for<'ctx> Fn(&'ctx Ctx, u32, u32) -> RequestFuture<'ctx, T>,
{
    let mut offset = 0;
    Box::pin(stream! {
        loop {
            // we create a reference to ctx here
            let page = req(&ctx, page_size, offset).await?;
            // due to `await`, the reference is not used again after this line
            offset += page.items.len() as u32;
            for item in page.items {
                // doesn't reference ctx, so yielding the items is fine
                yield Ok(item);
            }
            ...
        }
    })
}
```

> > and the result we get no longer references ctx so we can safely stream it to the consumer.
> 
> According to the function signature: `for<'ctx> Fn(&'ctx Ctx, u32, u32)`, the closure references to `ctx`, do you mean the result `RequestFuture` doesn't reference to `ctx`?

The function itself takes a reference and so does the `RequestFuture`, but when `RequestFuture` is `await`ed, we get a `ClientResult<T>` which does not reference `ctx`.

ramsayleung: I get your point, it makes sense:

```rust
pub fn paginate_with_ctx<'a, Ctx: 'a, T, Request>(...) -> Paginator<'a, ClientResult<T>>
where
    T: 'a + Unpin,
    Request: 'a + for<'ctx> Fn(&'ctx Ctx, u32, u32) -> RequestFuture<'ctx, T>,
{
    let mut offset = 0;
    Box::pin(stream! {
        loop {
            // we create a reference to ctx here. 
            // the `&ctx` is a shortlived reference, it's just like we put `ctx` into a bounded scope.
           //  ---- scope begins here ----
            let page = req(&ctx, page_size, offset).await?;
            // due to `await`, the reference is not used again after this line
           // ------ scope ends here -----
            offset += page.items.len() as u32;
            for item in page.items {
                // doesn't reference ctx, so yielding the items is fine
                yield Ok(item);
            }
            ...
        }
    })
}
```
 It's clear now :)

ramsayleung:
I think this PR is ready to merge, @eladyn would you like to wait for #361 merged to fix the Clippy error, or you could fix Clippy error in this PR, then #361 will not be a blocker?

eladyn:
@ramsayleung Everything should be fine now!

ramsayleung (approved):

ramsayleung:
Merged :)

Issue #349, opened 2022-07-25
bad request when trying to create playlist


fprasx:
**Describe the bug**
Whenever I try to use `user_playlist_create` I get `Error: http error: status code 400 Bad Request`. Other API calls work fine.

**To Reproduce**
Here is a minimum working example:
```rust
use anyhow::Result;
use rspotify::{prelude::*, scopes, AuthCodePkceSpotify, Config, Credentials, OAuth};

#[tokio::main]
async fn main() -> Result<()> {
    env_logger::init();

    let creds = Credentials::from_env().unwrap();

    let oauth = OAuth::from_env(scopes!(
        "playlist-read-collaborative",
        "playlist-read-private",
        "playlist-modify-public",
        "playlist-modify-private"
    ))
    .unwrap();

    let config = Config {
        token_cached: true,
        token_refreshing: true,
        ..Default::default()
    };

    let mut spotify = AuthCodePkceSpotify::with_config(creds, oauth, config);

    // The URL the user enters into the terminal
    let url = spotify.get_authorize_url(None).unwrap();

    match spotify.prompt_for_token(&url).await {
        Ok(_) => (),
        // If the token is too old, refresh it
        // The unwrap still panics sometime and I'm not sure why
        Err(_) => spotify.refresh_token().await.unwrap(),
    }

    let id = &spotify.me().await?.id;
    spotify
        .user_playlist_create(
            id,
            "this is a test",
            Some(true),
            Some(true),
            Some("this is a test"),
        )
        .await?;
    Ok(())
}
```
This panics with the error "Error: http error: status code 400 Bad Request". The API call `spotify.me()` works fine.

**Expected behavior**
I expected the playlist to be created (or have some other error like a network error).

**Log/Output data**
Running with `RUST_LOG=trace cargo run` produces the following:
```
[2022-07-25T13:11:39Z INFO  rspotify::auth_code_pkce] Building auth URL
[2022-07-25T13:11:39Z INFO  rspotify::auth_code_pkce] Generating PKCE codes
[2022-07-25T13:11:39Z INFO  rspotify::clients::oauth] Reading auth token cache
[2022-07-25T13:11:39Z INFO  rspotify::clients::base] Writing token cache
[...]

PR #355, opened 2022-07-29
add note for user_playlist_create endpoint

ramsayleung:
## Description

Add note:

> to create a collaborative playlist you must also set public to false

## Motivation and Context

#349 

## Dependencies 

None

## Type of change

## How has this been tested?

CI passes

It's unnecessary.


marioortizmanero (approved):
Thanks for the fix!

Issue #350, opened 2022-07-26
Make Spotify base URLs configurable for tests


jjmark15:
**Is your feature request related to a problem? Please describe.**
I want to be able to write tests against a local mock of Spotify. Spotify URLs are hard-coded in the library which makes this impractical.

e.g. https://github.com/ramsayleung/rspotify/blob/9197a908e9a61e20b47e22cac6cf9af75c6769b4/src/lib.rs#L188

**Describe the solution you'd like**
Make target base URLs configurable.

marioortizmanero:
I saw `aspotify` did something like that but I always wondered why it would be useful. So you've replicated the API? Would you be able to tell me more about it?

jjmark15:
Sure 👍🏻 . I generally prefer black-box functional (or integration depending on terminology) testing of an API. In order to do that I need to be able to make a real HTTP call to a downstream, which usually means using an http mock. Not sure if I've explained myself well enough there 😓

marioortizmanero:
I was wondering whether we should have these fields as a `String` or as a `&'static str`. It's not like anyone is going to modify the prefix URL at runtime, right?

PR #356, opened 2022-07-30
Add custom auth URLs and simplify `get`/etc methods

marioortizmanero:
## Description

This adds a configurable prefix URL for authentication.

I've also simplified the request methods in the base client. We don't really need `get` if it's only used in `endpoint_get`. We can also extend `post_form` to `auth_post`, which also prepends the auth URL for convenience, just like `endpoint_get`. Finally, I've renamed `endpoint_get` and similars to `api_get` for consistency and because it's clearer and much shorter.

## Motivation and Context

See #350 

## Dependencies 

None

## Type of change

## How has this been tested?

I've modified the `endpoint_url` test, and added another one for `auth_url`.

TODO

### DIFF
--- a/doc/uml/trait_hierarchy.plantuml
+++ b/doc/uml/trait_hierarchy.plantuml
@@ -9,7 +9,7 @@ abstract class BaseClient {
     ---
     ClientResult<()> auto_reauth()
     ClientResult<()> refresh_token()
-    String endpoint_url()
+    String api_url()
     String auth_url()
     ClientResult<Headers> auth_headers()
     ClientResult<()> write_token_cache()
--- a/doc/uml/trait_hierarchy.plantuml
+++ b/doc/uml/trait_hierarchy.plantuml
@@ -10,6 +10,7 @@ abstract class BaseClient {
     ClientResult<()> auto_reauth()
     ClientResult<()> refresh_token()
     String endpoint_url()
+    String auth_url()
     ClientResult<Headers> auth_headers()
     ClientResult<()> write_token_cache()
     ClientResult<Token> fetch_access_token()
--- a/src/clients/mod.rs
+++ b/src/clients/mod.rs
@@ -32,7 +32,7 @@ pub(crate) fn append_device_id(path: &str, device_id: Option<&str>) -> String {
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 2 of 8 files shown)


doc/uml/trait_hierarchy.plantuml
```diff
@@ -10,6 +10,7 @@ abstract class BaseClient {
     ClientResult<()> auto_reauth()
     ClientResult<()> refresh_token()
     String endpoint_url()
+    String auth_url()
```
marioortizmanero, line 13: @ramsayleung, can you generate the diagram again? Do you know if there's a way to do this automatically instead?

marioortizmanero:
Note that #375 should be merged before this.

doc/uml/trait_hierarchy.plantuml
ramsayleung: Yes, I can regenerate this diagram, but generating the diagram manually is a little clumsy, I would like to figure out how to generate this diagram automatically.

src/clients/mod.rs
```diff
@@ -55,19 +55,58 @@ mod test {
     }
 
     #[test]
-    fn test_endpoint_url() {
-        let spotify = ClientCredsSpotify::default();
+    fn test_api_url() {
+        let mut spotify = ClientCredsSpotify::default();
```
ramsayleung, line 59: Is it necessary to add the `mut` identifier to mark `spotify` mutable? It looks like the `api_url` doesn't require a mutable `Spotify`.

doc/uml/trait_hierarchy.plantuml
ramsayleung: I think we could use this GitHub Action to generate an image from `plantuml` file automatically: https://github.com/marketplace/actions/generate-plantuml

ramsayleung:
Everything looks good to me, except for the discussion about how to generate an image file from plantuml file automatically.

doc/uml/trait_hierarchy.plantuml
marioortizmanero: Great! Then you can ignore this comment, and we'll set it up in a separate PR.

ramsayleung (approved):

PR #359, opened 2022-09-05
add playlists

ctaggart:
I migrated some playlists to another account using https://www.musicbackuper.com/ to export the data to a json file. The loading failed, so I used this script instead. It deserializes the json file and then creates playlists and adds all the tracks.

I'm going to open this as a draft and then close it. I thought this may be a useful example issue for some.

Some challenges:
- The Spotify Web API to add a playlist doesn't respect `"public": false". All of the playlists end up being public.
https://community.spotify.com/t5/Spotify-for-Developers/Api-to-create-a-private-playlist-doesn-t-work/m-p/5418084#M5289
- It would have been better to reverse the list of playlists so that the oldest gets added first.
- Some tracks failed to add. I resorted to adding tracks one at a time because of this. The batches of 100 would fail if a single track was bad.

Thanks for the library! It was very useful.


ramsayleung:
It is great, thanks for your example and sharing :-)

Issue #324, opened 2022-06-11
Update `webapp` example


marioortizmanero:
**Is your feature request related to a problem? Please describe.**

The `webapp` example could definitely be improved.

**Describe the solution you'd like**

* Add token refreshing
* Use the new API properly (at some point we do `spotify.token = ...`, we can surely avoid that).
* Proper error handling? It has too many `unwrap` in my opinion.
* Add string interpolation where possible
* The Rocket framework isn't that active, unfortunately. It would surely be more helpful if it used Actix instead.
* Include it in the CI somehow so that this doesn't happen anymore.

**Describe alternatives you've considered**

I've fixed a couple things in #305, but it was out of scope.

ramsayleung:
I totally agree with your point, the `webapp` example is a little stale, we should refactor this example with the new API, to make it polished and robust.

Except for this point:
+ The Rocket framework isn't that active, unfortunately. It would surely be more helpful if it used Actix instead.

I am not sure if we could be beneficial from switching to `Actix`. Comparing the recent PR history of Rocket with Actix,  I think there are roughly active (or inactive?).

marioortizmanero:
> The Rocket framework isn't that active, unfortunately. It would surely be more helpful if it used Actix instead.

Fair enough. I was worried because v0.5.0 hasn't been released yet, but it does seem like the repo is active.

> And we should start to refactor webapp example after https://github.com/ramsayleung/rspotify/pull/305 merged, and honestly speaking, https://github.com/ramsayleung/rspotify/pull/305 goes a little far so that I need times to get involved in.

Sorry, I didn't quite understand that. What do you mean?

ramsayleung:
> Sorry, I didn't quite understand that. What do you mean?

Never mind, I just think the work of refactoring the `webapp` example should wait until #305 get merged.

graves501:
Since #305 already got merged, I guess the `webapp` is ready for some polish?

I tried out the `webapp` example and I noticed that I am not able to fetch the top artists...

```rust
#[get("/topartists")]
fn top_artists(cookies: Cookies) -> AppResponse {
    let mut spotify = init_spotify(&cookies);
    if !spotify.config.cache_path.exists() {
        return AppResponse::Redirect(Redirect::to("/"));
    }

    let token = spotify.read_token_cache(false).unwrap();
    spotify.token = Arc::new(Mutex::new(token));

    let top_artists = spotify
        .current_user_top_artists(Some(&TimeRange::LongTerm))
[...]

PR #361, opened 2022-09-20
Update webapp example

ramsayleung:
## Description

Features:

- [x] Add token refreshing
- [x] Use the new API properly (at some point we do `spotify.token = ...,` we can surely avoid that).
- [x] Proper error handling? It has too many unwrap in my opinion.
- [x] Add string interpolation where possible
- [x] Include it in the CI somehow so that this doesn't happen anymore.

## Motivation and Context

#324 

## Dependencies 

+ log
+ env_log
+ cookie

## Type of change

Not yet.

### DIFF
--- a/.github/workflows/webapp_ci.yml
+++ b/.github/workflows/webapp_ci.yml
@@ -3,14 +3,10 @@ on:
   push:
     branches: master
 
-defaults:
-  run:
-    working-directory: examples/webapp
-
 name: WebApp Continuous Integration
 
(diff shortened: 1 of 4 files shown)


marioortizmanero:
Thanks for the contribution and sorry for the late PR! Can confirm that this works wonderfully and fixes #324. We should first merge #374 so that CI passes all the checks.

marioortizmanero (approved):

PR #364, opened 2022-10-16
Added a builder like struct for the search function (Related with issue )

### LINKED ISSUE #354: Type-safe searching

The [search endpoint](https://developer.spotify.com/documentation/web-api/reference/#/operations/search) requires specifying parameters and similars. We currently only require a `q` string, and the user has to build it themselves, which is pretty barebones.

It'd be great to have a builder pattern or any data structure that would make said step easier.

Leaving it as it is, but usability isn't good.

**Additional context**

See the troubles found in #351

GartoxFR:
## Description

These changes add a struct to build well formatted query to pass to the search method (fixes #354).
These are not breaking changes as you can still pass a simple String to the method.
It can still be improve as it doesn't prevent users to pass wrong filter with wrong SearchType (give a track title to search an album for example, which is not supporte by Spotify API).

## Motivation and Context

Make writing searches easier

## Dependencies 

Adding strum to the main project

## Type of change

## How has this been tested?

I've modified one test case to use the struct instead of a basic &str. As the old test still passes, these changes are non breaking.

 I also wrote a simple example which printed the String generated by the struct to verify that it was right with its inputs. This file is included in this PR but will probably be removed if it is accepted.


Cargo.toml
```diff
@@ -41,6 +41,7 @@ log = "0.4.14"
 maybe-async = "0.2.6"
 serde = { version = "1.0.130", default-features = false }
 serde_json = "1.0.67"
+strum = { version = "0.24.0", features = ["derive"] }
 sha2 = "0.10.0"
```
ramsayleung, line 45: Is it necessary to impose the `strum` crate? We are cautious about including a new crate, it will increase compile time.

src/search/mod.rs
```diff
@@ -0,0 +1,96 @@
+use std::collections::HashMap;
+
+use strum::Display;
+
+#[derive(Debug, Display, PartialEq, Eq, Hash)]
+#[strum(serialize_all = "snake_case")]
```
ramsayleung, line 6: I think the `SearchFilter` enum should be put into the `rspotify-model` crate, then we don't need to impose a new crate and `SearchFilter` is more like a model.

examples/search_query.rs
```diff
@@ -0,0 +1,13 @@
...
+#[tokio::main]
+async fn main() {
+    let query: String = SearchQuery::default()
+        .any("any")
+        .artist("Lisa")
+        .track("Demon slayer")
+        .album("Another Album")
+        .into();
+
+    println!("{}", query);
```
ramsayleung, line 12: I think it's not a full example, right? The `SearchQuery` should combine with the `search`  endpoint?

src/search/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+    TagNew,
+    Isrc,
+    Genre,
+}
+
+#[derive(Debug, Default)]
+pub struct SearchQuery {
+    no_filter_query: String,
+    query_map: HashMap<SearchFilter, String>,
+}
```
ramsayleung, line 25: According to the documentation, there is a restriction between filter and type
> You can narrow down your search using field filters. The available filters are `album`, `artist`, `track`, `year`, `upc`, `tag:hipster`, `tag:new`, `isrc`, and `genre`. Each field filter only applies to certain result types.

> The artist and year filters can be used while searching albums, artists and tracks. You can filter on a single year or a range (e.g. 1955-1960).
The `album` filter can be used while searching albums and tracks.
The `genre` filter can be used while searching artists and tracks.
The `isrc` and track filters can be used while searching tracks.
The `upc`, tag:new and tag:hipster filters can only be used while searching albums. The `tag:new` filter will return albums released in the past two weeks, and `tag:hipster` can be used to return only albums with the lowest 10% popularity.

Thus, I think we should take the  `search type` into account when designing the `SearchQuery` Model.

src/search/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    pub fn genre<T: Into<String>>(&mut self, str: T) -> &mut Self {
+        self.query_map.insert(SearchFilter::Genre, str.into());
+        self
+    }
+}
+
+impl From<&mut SearchQuery> for String {
+    fn from(val: &mut SearchQuery) -> Self {
```
ramsayleung, line 80: We have not changed the `val` variable, so we could remove the `mut`.

src/search/mod.rs
GartoxFR: I thought about storing the SearchType into the SearchQuery but that means the search method would not need the SearchType anymore and it would be breaking change, I'll try to design this with a method I name differently for to see how it could look

Cargo.toml
GartoxFR: Will it really increase compile time ? As strum is built while building models which are built before the main crate. Anyway I think moving SearchFilter to the models is a good idea and will solve this problem

src/search/mod.rs
GartoxFR: I don't see any filters compatible with Shows, Episodes and Playlists. They don't have any advanced filtering keyword ?

GartoxFR:
I tried an other approach to ensure type safety. Please tell me what you think about this before I try to implement testing and more example around it

src/search/mod.rs
ramsayleung: > 

I don't see any filters compatible with Shows, Episodes and Playlists. They don't have any advanced filtering keyword ?

I think we need to dig deep into the Spotify document to check this, it seems not point out in the Spotify document

ramsayleung:
After rereading the Spotify's document, I think its explanation about the restriction between `filter` and `result type` is ambiguous, so I think it's better to leave `filter` and `result type` alone, roll back to your original design. 

Just adding a builder for the `query string`, and we are unnecessary to include the `search type` into the builder.

GartoxFR:
I reverted my previous changes to go back to the builder pattern.
I'm not used to writing test so can you give me ideas on how to write test for this ?
I can't just call the method into and compare the result to an expected query string because the order of the keys in the query string is undefined and can change

ramsayleung:
In my opinion, there are two ways to test the `SearchQuery` builder.
> I can't just call the method into and compare the result to an expected query string because the order of the keys in the query string is undefined and can change
1. Just making the order of the keys determined and predictable by replacing `HashMap` with [`BTreeMap`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeMap.html) which is a sorted map by itself. Then the complexity of querying if a key is in the map will reduce from `O(1)` to` O(LogN)`, but I think it's acceptable.
2. Test the query string word by word, for example:
```rust
let query_string = SearchQuery::default().album("arrival").artist("abba").tag_new("test");

// the query_string could be:
// 1. `ablum:arrival artist:abba test`
// 2. `artist:abba ablum:arrival test`
// 3. `test artist:abba ablum:arrival`
// 4. ...

// we could split query string by whitespace and assert the splited strings contain `ablum:arrival` `artist:abba` `test`
let query_words: Vec<String> = query_string.split(' ').collect();
assert!(query_words.contains('ablum:arrival'));
assert!(query_words.contains('artist:abba'));
// ...
```

GartoxFR:
I finally chose the BTreeMap because it's a bit tricky to split correctly the query as there can be spaces inside album, track..

src/search/mod.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Default)]
+pub struct SearchQuery<'a> {
+    no_filter_query: &'a str,
+    query_map: BTreeMap<SearchFilter, &'a str>,
+}
+
+impl<'a> SearchQuery<'a> {
+    /// Basic filter where the given string can be anything
+    pub fn any(&mut self, str: &'a str) -> &mut Self {
+        self.no_filter_query = str;
```
marioortizmanero, line 31: To me, it'd be more intuitive if calling `any` multiple times appended its contents, instead of replace them.

marioortizmanero (commented):
Awesome idea, thanks for the contribution!!

GartoxFR:
I'm a bit confused about what did just happen in the pipeline. It seems it complained about some code which is inside the documentation. I never really worked with rust doc so what happened ?

ramsayleung:
The pipeline complained about that it's unable to find the definition of `SearchQuery`, so you should `import` the `SearchQuery` before using it:

```rust
/// Exemple
/// ```rust
/// use SearchQuery;
/// SearchQuery::default()
///     .any("foo")
///     .album("bar")
/// // Filter on album containing "bar" and anything containing "foo"
/// ```
```

PR #366, opened 2022-10-27
Replace std::time::Duration with chrono::Duration to support negative duration

ramsayleung:
## Description

https://github.com/ramsayleung/rspotify/issues/363

## Motivation and Context

https://github.com/ramsayleung/rspotify/issues/363

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## How has this been tested?

All existing tests pass.

- [x] Test A: test_resume_point_negative

## Is this change properly documented?

Yes

PR #375, opened 2022-12-27
More `chrono::Duration` usage in the API

marioortizmanero:
## Description

This uses more `chrono::Duration` in the API, as opposed to the more basic `i32` and `u32`. It also renames some fields to `position` from `position_ms`

## Motivation and Context

I noticed this in #366; using `chrono::Duration` is more versatile and idiomatic. Not sure if I missed any cases.

## Dependencies 

Nothing new.

## Type of change

## How has this been tested?

CI still passes

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 - ([#331](https://github.com/ramsayleung/rspotify/pull/331)) `Market` is now `Copy`
 - ([#366](https://github.com/ramsayleung/rspotify/pull/366)) Replace all `std::time::Duration` with `chrono::Duration` to support negative duration.
-- ([#TODO](https://github.com/ramsayleung/rspotify/pull/TODO)) We now use
+- ([#375](https://github.com/ramsayleung/rspotify/pull/375)) We now use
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 - ([#331](https://github.com/ramsayleung/rspotify/pull/331)) `Market` is now `Copy`
 - ([#366](https://github.com/ramsayleung/rspotify/pull/366)) Replace all `std::time::Duration` with `chrono::Duration` to support negative duration.
+- ([#375](https://github.com/ramsayleung/rspotify/pull/375)) We now use
```
ramsayleung, line 5: Since `v0.11.6` has released, these changes should be record for the `v.0.11.7`

ramsayleung (approved):

CHANGELOG.md
marioortizmanero: Fixed. Shall I merge the PR, then?

ramsayleung: Sure, I have merged this PR

jirutka:
This is a breaking change, so the major version should have been bumped. Introducing it in a patch version is really not good.

ramsayleung:
Yes, it should be. I would like to release a new version

PR #379, opened 2022-12-29
Much simpler interface based on webapp example

### LINKED ISSUE #324: Update `webapp` example

marioortizmanero:
## Description

This PR simplifies `rspotify`'s usage based on the webapp example. I realized that we were really handing the token cache ourselves, instead of being automatic. We had lots of unnecessary boilerplate, which I've now simplified thanks to a refactored auth flow:

- I've moved `read_token_cache` into `BaseClient` and generalized it to all kinds of clients through the method `is_token_valid`. Note that valid doesn't mean that it's ready to be used, just that it matches the flow, and that it could be used to reauthenticate (i.e., even if expired, we have a refresh token).
- I've modified `write_token_cache` so that it needs to have the `token` passed, instead of only using `self.token`. Not only does this avoid unnecessary locks in some places (and weird code because we don't want to lock inside a lock), but also it's more consistent with the way `read_cache_token` works.
- We now check the cache file once at the first request in the client. If it's valid, we set it as the client's token. Note that it might be valid but not refreshed, so the user will need to make sure both `token_cached` and `token_refreshing` are enabled. Otherwise, they'll have to take care of that themselves.

## Motivation and Context

This closes #324

## Dependencies 

`parking_lot` makes `token`'s usage simpler by simply not dealing with lock poisoning manually. We can avoid `unwrap` for every lock we have. Furthermore, I've removed the async variant, as many sources indicate that a regular mutex is usually better anyway.

## Type of change

## How has this been tested?
Ran it locally
Yes.


marioortizmanero:
This is not ready to be merged yet, but let me know your thoughts, @ramsayleung. This is something I should've done earlier, cached tokens were all over the place; now it's truly automatic.

src/clients/base.rs
```diff
@@ -87,14 +112,16 @@ where
     /// automatic reauthentication takes place, if enabled.
     #[doc(hidden)]
     async fn auth_headers(&self) -> Headers {
+        self.auto_cache()
+            .await
+            .expect("Failed to read or write to cache automatically");
+
```
marioortizmanero, line 118: Most important line in the PR: reading from the cache in the first request.

PR #388, opened 2023-02-03
Remove rustc-serialize feature from chrono

### LINKED ISSUE #387: Using rustc-serialize
The crate rspotify and rspotify-model are depending on chrono, and the Cargo.toml of both craters are enabling the optional rustc-serialize feature of chrono, dragging it into the dependency tree.  Seems that rustc-serialize eventually rustc will refuse to compile it:

```
warning: the following packages contain code that will be rejected by a future version of Rust: rustc-serialize v0.3.24
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

```
The package `rustc-serialize v0.3.24` currently triggers the following future incompatibility lints:
> warning: impl method assumes more implied bounds than the corresponding trait method
>     --> /home/aplanas/.cargo/registry/src/github.com-1ecc6299db9ec823/rustc-serialize-0.3.24/src/serialize.rs:1155:41
>      |
> 1155 |     fn decode<D: Decoder>(d: &mut D) -> Result<Cow<'static, T>, D::Error> {
>      |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: replace this type to make the impl signature compatible: `Result<Cow<'a, T>, <D as serialize::Decoder>::Error>`
>      |
>      = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
>      = note: for more information, see issue #105572 <https://github.com/rust-lang/rust/issues/105572>
>      = note: `#[allow(implied_bounds_entailment)]` on by default
>
```

IIUC this means this crate will not be possible to compile in the future.

Maybe the fix is to replace https://github.com/ramsayleung/rspotify/blob/master/Cargo.toml#L36 with

    chrono = { version = "0.4.19", features = ["serde"] }

And a similar change in rspotify-model Cargo.toml

aplanas:
Because rustc-serialize is considered insecured, and is using constructor that will be not compiled in future rustc, chrono dropped rustc-serialize in the main branch (future 0.5 release)

This patch remove this feature, to avoid the current compiler warning and prepare the work for the new chrono release.

Fix #387

Signed-off-by: Alberto Planas <aplanas@suse.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ rspotify-http = { path = "rspotify-http", version = "0.11.6", default-features =
 async-stream = { version = "0.3.2", optional = true }
 async-trait = { version = "0.1.51", optional = true }
 base64 = "0.20.0"
-chrono = { version = "0.4.19", features = ["serde", "rustc-serialize"] }
+chrono = { version = "0.4.19", features = ["serde"] }
 dotenv = { version = "0.15.0", optional = true }
 futures = { version = "0.3.17", optional = true }
 getrandom = "0.2.3"
--- a/rspotify-http/src/common.rs
+++ b/rspotify-http/src/common.rs
@@ -37,11 +37,11 @@ pub trait BaseHttpClient: Send + Default + Clone + fmt::Debug {
         payload: &Value,
     ) -> Result<String, Self::Error>;
 
-    async fn post_form<'a>(
+    async fn post_form(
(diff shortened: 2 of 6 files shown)


ramsayleung (approved):

ramsayleung:
Thanks for your great work :)

PR #391, opened 2023-02-25
Make Paginator Send to enable scheduling onto a multithreaded runtime

darkwisebear:
## Description

Paginator is in fact a boxed trait object. However, it doesn't implement Send which is a requirement when scheduling onto a async multithreaded scheduler.

## Motivation and Context

Using a Paginator which is not Send will also make the enclosing Future not Send. Therefore , when using tokio with the default scheduler, this will only work in the main thread.

## Dependencies 

None.

## Type of change

## How has this been tested?

The first commit adds a test that sets `Sync`, which multithreaded async schedulers usually have. This test will not compile without modifying the code, just like if you try to schedule a future using a Paginator without a Send bound.

- Using a Future enclosing an (async) Paginator, scheduled onto the tokio multithreaded scheduler.

Since this change only relaxes requirements for the user of the Paginator, no documentation change was made.

### DIFF
--- a/src/clients/pagination/stream.rs
+++ b/src/clients/pagination/stream.rs
@@ -58,3 +58,40 @@ where
         }
     })
 }
+
+#[cfg(test)]
+mod test {
+    use super::paginate;
+    use crate::model::Page;
+    use futures::{future, StreamExt};
+    use std::future::Future;
+
+    fn schedule_future<'a, F>(fut: F)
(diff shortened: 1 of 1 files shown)


ramsayleung (approved):

ramsayleung:
Great work! Thanks for your contribution.

PR #394, opened 2023-03-09
enable / standardize client request timeouts

eladyn:
## Description

This sets the same timeout duration for `reqwest` and `ureq` http clients.

## Motivation and Context

https://github.com/ramsayleung/rspotify/blob/6d304de632a84f79c35cb9ff4a0e66be238cf205/rspotify-model/src/auth.rs#L79-L80

This mentions a maximum request time of 10 seconds, but AFAICT, these timeouts aren't specified anywhere. By default, `reqwest` does not have a timeout at all, `ureq` has a 30 second `connect` timeout.

## Type of change

## How has this been tested?

The provided non-modifying tests.

I hope so. :) Please point me to additional documentation changes that are necessary, if there are any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - ([#384](https://github.com/ramsayleung/rspotify/pull/384)) Add STB alias for Stb device type, fix for `json parse error: unknown variant STB`.
 - ([#386](https://github.com/ramsayleung/rspotify/pull/386)) Support `BaseClient::artist_albums` with zero or more `AlbumType`.
 
+**Bugfixes**
+- ([#394](https://github.com/ramsayleung/rspotify/pull/394)) Set a common 10 second timeout for both http clients.
+
 ## 0.11.6 (2022.12.14)
 
 - ([#331](https://github.com/ramsayleung/rspotify/pull/331)) `Market` is now `Copy`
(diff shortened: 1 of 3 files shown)


ramsayleung:
> This mentions a maximum request time of 10 seconds, but AFAICT, these timeouts aren't specified anywhere. By default, `reqwest` does not have a timeout at all, `ureq` has a 30 second `connect` timeout.

Thanks for your fix, you make a great point. These two HTTP client do has a different default timeout setting.

But I want to clarify the code you are referring, I am not sure whether you are misunderstanding the comment.

https://github.com/ramsayleung/rspotify/blob/6d304de632a84f79c35cb9ff4a0e66be238cf205/rspotify-model/src/auth.rs#L79-L80

The 10 seconds margin in the comment, means the margin of token expiration, instead of the HTTP `connect` timeout.

For example, if the expiration time is `2023-03-10 00:00:10`, the auth flow will check and assert that it's expired if the current time is later than `2023-03-10 00:00:00`, the 10 seconds is a margin.

From this point, it makes sense to set the HTTP `connect` timeout to 10 seconds :)

eladyn:
> The 10 seconds margin in the comment, means the margin of token expiration, instead of the HTTP `connect` timeout.

Yeah, that makes sense.

> From this point, it makes sense to set the HTTP `connect` timeout to 10 seconds :)

Do you mean that rather than applying a general timeout (which includes the connect, write and read phase), we should only be setting a connect timeout? Or is it fine the way I did it?

ramsayleung:
> Or is it fine the way I did it?

Your approach is great. However, I would like to clarify any potential misunderstandings :)

ramsayleung (approved):

Issue #400, opened 2023-04-19
Provide a user-settable callback for when tokens are updated


PaulOlteanu:
### Is your feature request related to a problem? Please describe
I'm creating a web-app that stores users' tokens in a database. After performing a Spotify request, I don't have a way to know if the token was refreshed, and therefore have no way to store the updated tokens in my database.

### Describe the solution you'd like
I'd like a config option on the clients where I could provide a callback function that could be called when the token is refreshed, so that I can save it to my database.

I feel that this solution is much more "library-like" than the current one that optionally saves the token to a file.

### Describe alternatives you've considered
One alternative that I've considered (that I don't think is very good) is to set a field on the client when the token is changed, that can be unset by a user once they've handled the new token.

### Additional context
I could try to look into implementing this if we think it's a good thing to add to rspotify

PaulOlteanu:
I think it may be better to have this be a breaking change rather than try to keep the previous behaviour of reading/writing the token.

To me it seem strange to have that kind of stuff happening from a library (and in general I personally feel that there's other things that don't really belong as part of a library such as the env-file feature - I think the end user should be doing that themselves if they need envfile support).

Maybe that should be part of a different discussion on a separate issue.

PaulOlteanu:
I've opened a draft PR in case anyone has some early feedback. I'm going to look into trying to use generics rather than trait objects to see how that impacts the ergonomics for the user. It would mean the clients would be generic over the functions the user provides as callbacks.

If anyone has opinions on either option (`dyn Fn` vs `impl Fn`) I'd love to hear them

ramsayleung:
> I think it may be better to have this be a breaking change rather than try to keep the previous behaviour of reading/writing the token.

> To me it seem strange to have that kind of stuff happening from a library (and in general I personally feel that there's other things that don't really belong as part of a library such as the env-file feature - I think the end user should be doing that themselves if they need envfile support).

I agree with your point, it's better to be a breaking change rather than try to keep the previous behavior of reading/writing the token. But in order to minimize the impact of making a breaking change for user/developer, I prefer to provide a feature named `token-file` or other names, to provide a feature to user to keep everything unchanged, which is disabled by default.

PaulOlteanu:
Yeah, apologies for not finishing this up myself sooner.

Anyway I found that due to the way generics are handled we wouldn't be able to make the clients generic over the callback type because Rust would require the function to implement Debug even if we store it in some wrapper type.

[...]

PR #418, opened 2023-05-27
Add token fetch callback

ramsayleung:
## Description

Add a user-customized callback function support which is invoked whenever client succeeds to request or refresh a token, so it provides users an opportunity to save token into file, store into database etc.

## Motivation and Context

#400 

## Dependencies 

None

## Type of change

## How has this been tested?

- [x] `with_token_callback_fn.rs`


PaulOlteanu:
I have some tests written as well https://github.com/PaulOlteanu/rspotify/commit/2d38341478bfa950225a2fbd8e6e5d6ddcb4c5ac if you'd like to add those

ramsayleung:
Thanks for your contribution, I would like to add your test, but I refine the callback function based your code. 

The callback function is a field of `Config` struct instead of `SpotifyClient`, so it's unnecessary to use `Arc<Mutex>` to sync between different thread.

`TokenCallback` should only be invoked when a token is gained successfully, so it's unnecessary to use `Fn(Option<Token>`, `Fn(Token)` is more appropriate. 

Furthermore, the callback function might throw an error, so the return value of the callback function should be `Result`

ramsayleung:
Sure, thanks for your contribution in advance.

Issue #419, opened 2023-05-30
0.11.17 breaks PKCE auth


jacobmichels:
**Describe the bug**
`prompt_for_token` panics after inputting the redirect url from the browser. The logs reveals that the http client sent a request to /api/token but received a 400 response. The response body states that the code_verifier was not correct. This issue is not present in 0.11.6 or 0.11.5.

**To Reproduce**
1. Checkout rspotify v0.11.7
2. Run the auth_code_pkce example. 
3. Observe a panic after pasting in the redirect url

**Expected behavior**
`prompt_for_token` correctly formulates the request for the access token as it does before v0.11.7

**Log/Output data**
``` 
$ RSPOTIFY_CLIENT_ID=... RSPOTIFY_CLIENT_SECRET=... RSPOTIFY_REDIRECT_URI='http://localhost:8000/callback' cargo run --features="env-file cli client-reqwest" --example auth_code_pkce
     Finished dev [unoptimized + debuginfo] target(s) in 0.08s
     Running `target/debug/examples/auth_code_pkce`
Opened https://accounts.spotify.com/authorize?client_id=...&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A8000%2Fcallback&code_challenge_method=S256&code_challenge=lwVEozlFwPIHbPZtH8ohjISa%2BlW%2BU%2BTER9nt1o5SYBU%3D&state=7XmghTMdx6xQZXW7&scope=user-read-playback-state in your browser.
Please enter the URL you were redirected to: 
http://localhost:8000/callback?code=AQCJPjYPRU6_A3fnQkI5osldxE9En9uDaweca8LpXc_oOr0P_8bd0ADz3cANdohdkLmcMebzGk9qx8WbVPvExmHNfJe3tkNCdM7CCeuQdg6D0kqMCrqHgEzNNSuNfWB4r5O6PDdst8MbSP2vffKmriJLZtCgt-5-wDyddmq0ozMushoM_2T0Z1NRSz2V38iNnMS1BY6uVYuTcvR3zPJA3uF3Iqv3z81iWUuIM8lIFIKyhXliTNIDw56pBbRZOT9JnZWlfuxUhAXq2xxIpzvKgQ&state=7XmghTMdx6xQZXW7
[...]

PR #420, opened 2023-05-30
Fix pkce

jacobmichels:
## Description

This PR fixes the PKCE auth flow panicking during `prompt_for_token`. The issue was the challenge code was being base64 encoded in a url-unsafe manner. This PR simply changes the base64 encoding logic to be url safe.

I also bumped the base64 crate version and updated lib.rs to avoid using a deprecated base64 encode API.

#419

## Motivation and Context

I wanted to use PKCE auth.

## Dependencies 

No dependencies added. base64 bumped from 0.20.0 to 0.21.2

## Type of change

## How has this been tested?

I didn't do any automated testing, I just ran the auth_code_pkce example to check if my changes worked. On version 0.11.7 this example panics. With my patch, the example functions as expected.

I added an entry to the changelog. I don't believe further documentation is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 **Breaking changes**
 - ([#409](https://github.com/ramsayleung/rspotify/pull/409)) Change type of `position` parameter in `playlist_add_items` endpoint from `Opinion<Duration>` to `Opinion<u32>`
 
+**Bugfixes**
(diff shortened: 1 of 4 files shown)


ramsayleung:
I think an existing unit test failed because of the base64 modification.

jacobmichels:
Seen that, just pushed a commit to fix it. I'm not sure what base64 encoding alphabet is expected for basic auth, whether it should be standard or url safe. I'm leaning towards standard

ramsayleung (approved):

PR #434, opened 2023-08-14
Bug Fix: Set refresh token to new access token before `callback_fn` is called

JoeyEamigh:
## Description

`callback_fn` was not returning the refresh token in `AuthCodeSpotify`, so I simply moved the `token.refresh_token = Some(refresh_token.to_string());` above the `callback_fn` directive to solve this problem.

## Motivation and Context

Refresh token was not being returned after `refetch_token` was called.

## Type of change

## How has this been tested?

I used it in my project and it does work. It is a two line 

I figure this can be done once the PR is in.

Thanks for the great library!


ramsayleung (approved):

ramsayleung:

Issue #438, opened 2023-08-27
Is Add Custom Playlist Cover Image request implemented ?


tomkimsour:
**Is your feature request related to a problem? Please describe.**
I want to add a image to a created playlist

**Describe the solution you'd like**
A function like ```current_user_playlist_cover_update()``` would be convinient to use.
API reference : https://developer.spotify.com/documentation/web-api/reference/upload-custom-playlist-cover

tomkimsour:
I am willing to create a PR for this. I am just not sure if that exist already or not.

ramsayleung:
This API is not implemented yet, and PR is definitely welcome :)

Issue #377, opened 2022-12-29
Implement `image` endpoints for playlists


marioortizmanero:
**Is your feature request related to a problem? Please describe.**

We're missing the endpoints to read and set a playlist's cover image:

- https://developer.spotify.com/documentation/web-api/reference/#/operations/get-playlist-cover
- https://developer.spotify.com/documentation/web-api/reference/#/operations/upload-custom-playlist-cover

**Describe the solution you'd like**

Implement these endpoints and update the model.

**Additional context**

This was originally brought up in #16.

herberthudson:
Hi,
for the upload I still couldn't test it, I just wrote the code, and since I'm still new to rust, I couldn't think of how I could write tests to validate the code. If you prefer, I can send it in a separate branch if you want to review the code. In any case, when I have more time, I'll come back to it.

Issue #154, opened 2020-11-16
Testing OAuth endpoints


ramsayleung:
It's a little bit awkward that most endpoints implemented in `rspotify` which require `oauth` token to go haven't have some related unit tests. 

Since most tests will depend on Spotify Web API, and have to obtain the `oauth` token before sending request to Spotify server. The problem there is no a proper way to get `oauth` token, it has to open a browser and sign in with specific scopes, and now the process can't be automated. 

```rust
#[maybe_async]
#[maybe_async_test]
#[ignore] // we have to ignore those tests for the reason discussed above.
async fn test_current_user_playing_track() {
    oauth_client()
        .await
        .current_user_playing_track()
        .await
        .unwrap();
}

```

As everything goes without saying, tests are essential, and we should figure out a proper way to add test for endpoints.

marioortizmanero:
The main blocking issue is that we don't have a premium account to run these tests on. I tried to request tokens for library maintainers, here's the post in their forum for developers: https://community.spotify.com/t5/Spotify-for-Developers/Dummy-premium-accounts-for-library-developers/m-p/5046929. For now you can press Like and leave a comment with your thoughts, but it seems to be somewhat stale. We should notify other maintainers for spotify libraries to support this issue and get some traction.

Before we run any tests with oauth though we have to make sure they don't modify the user's data (for example first testing following users and then unfollowing them).

marioortizmanero:
Additionally, it could be a good idea to have a GitHub action with a cronjob that periodically (say, once a day) checks that the endpoints work. This would let us know about changes in the Spotify API very early and give us more time to apply them. Here's how tekore does it.

thecakeisalie25:
Could you not use a [self-hosted runner](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners) and your own personal spotify account? While it isn't ideal, the spotify devs don't exactly seem receptive to the issue. 

This would keep the credentials private, since nobody else has access to the .env file or your token. As far as not modifying the account, are there any API calls that don't have an inverse? You could test the playlist creation endpoint, which creates a playlist, then test the get user playlists endpoint, which would find that playlist (and ignore all the others), then test the delete user playlists endpoint, which would delete that playlist. Overall, this has no net impact on your playlists, since it creates and then deletes one immediately after. I'm unaware of any endpoints that don't have some inverse operation like that (or have any impact at all).

Also, what are the restrictions on a free account? No API access at all?

[...]

PR #439, opened 2023-08-30
Add image cover

tomkimsour:
## Description

Please include a summary of the change and which issue is fixed
Getter and setter for playlist cover are missing as mentioned here #377 and #438

## Motivation and Context

Please also include relevant motivation and context. 

## Dependencies 

None

## Type of change

## How has this been tested?

I am yet trying to test both function under test_with_oauth.rs by creating a playlist, adding a cover, retrieving it and comparing both.
However I need guidance on how to test it as I am not able to figure out how to oauth for the tests. I am not sure what are the variable values required in the .env and even if I put a wrong value, it doesn't seem like the value is being read.

yes

### DIFF
--- a/src/clients/base.rs
+++ b/src/clients/base.rs
@@ -574,6 +574,24 @@ where
         convert_result(&result)
     }
 
+    /// Get cover image of a playlist.
+    ///
+    /// Parameters:
+    /// - playlist_id - the playlist ID, URI or URL
+    /// [reference](https://developer.spotify.com/documentation/web-api/reference/get-playlist-cover)
+    async fn playlist_cover_image(
+        &self,
+        playlist_id: PlaylistId<'_>,
+    ) -> ClientResult<Option<Image>> {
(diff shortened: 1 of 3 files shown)


ramsayleung:
You could take the [`current_playing`](https://github.com/ramsayleung/rspotify/blob/master/examples/auth_code.rs) endpoint for example, test your new endpoint by 
1. Add an example file
2. Add an [entry](https://github.com/ramsayleung/rspotify/blob/master/Cargo.toml#L94) in cargo.toml
3. Run your example with `cargo`: `cargo run --example your-new-example --features env-file,cli,client-reqwest`, the cli will prompt you to authenticate and get the auth code.

ramsayleung:
The example acts like an end-to-end test for the new endpoint, because as we discussed in this issue: https://github.com/ramsayleung/rspotify/issues/154#issuecomment-1699115854, there is no workable approach for us to run unit test.

PR #444, opened 2023-11-04
[docs] Add comment explaining required deps

beyarkay:
This is a no-code commit. The `pagination_async` example doesn't explicitly say that the `futures` and `futures_util` crates are required, so I had to do a lil' hunting to figure out what crates I was missing when I tried to copy-paste the code.

This commit just adds a comment explaining how to add the required dependencies.


ramsayleung:

Issue #464, opened 2024-03-08
SimplifiedAlbum added onto SimplifiedTrack Struct


channingbabb:
**Describe the bug**
I believe that the SimplifiedTrack struct should have the SimplifiedAlbum per [Recommendations on Spotify Developers API](https://developer.spotify.com/documentation/web-api/reference/get-recommendations).

**Expected behavior**
Being able to access album information when getting track recommendations.  

**Additional context**
Here's the member that I added to the SimplifiedTrack struct that worked properly:
`pub album: SimplifiedAlbum`

Which is the same as the FullTrack album member.

PR #465, opened 2024-03-08
Add SimplifiedAlbum to SimplifiedTrack.

channingbabb:
## Description

This adds SimplifiedAlbum member to the SimplifiedTrack struct.

## Motivation and Context

Found this issue through the recommendations API.
Fixes: https://github.com/ramsayleung/rspotify/issues/464#issue-2176647918

## Type of change

**- [X] Bug fix (non-breaking change which fixes an issue)**

## How has this been tested?
Using this function:

    pub async fn get_user_suggestions(spotify_session: AuthCodeSpotify) -> Recommendations {
        let seed_artists = [ArtistId::from_id("4NHQUGzhtTLFvgF5SZesLK").unwrap()];

        let attributes = [RecommendationsAttribute::MinEnergy(0.4)];

        let recommendations: Recommendations = spotify_session.recommendations(
            attributes,
            Some(seed_artists),
            None::<Vec<&str>>,
            Some(None),
            Some(Market::Country(Country::UnitedStates)),
            Some(100)
        )
            .await
            .unwrap();

        recommendations
    }
    
You were not able to see Album information prior to the member addition.

### DIFF
--- a/rspotify-model/src/track.rs
+++ b/rspotify-model/src/track.rs
@@ -59,6 +59,7 @@ pub struct FullTracks {
 /// relinking is applied.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct SimplifiedTrack {
+    pub album: SimplifiedAlbum,
     pub artists: Vec<SimplifiedArtist>,
     pub available_markets: Option<Vec<String>>,
     pub disc_number: i32,
--- a/rspotify-model/src/track.rs
+++ b/rspotify-model/src/track.rs
@@ -59,7 +59,8 @@ pub struct FullTracks {
 /// relinking is applied.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct SimplifiedTrack {
-    pub album: SimplifiedAlbum,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub album: Option<SimplifiedAlbum>,
     pub artists: Vec<SimplifiedArtist>,
     pub available_markets: Option<Vec<String>>,
     pub disc_number: i32,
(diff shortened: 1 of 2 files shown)


ramsayleung:
Thanks for the contribution, the `test_simplified_track` test failed because you haven't updated the associated test yet.

channingbabb:
> Thanks for the contribution, the `test_simplified_track` test failed because you haven't updated the associated test yet.

I believe I updated it correctly now!  Sorry about that!

ramsayleung:
The CI workflows still failed.

rspotify-model/src/track.rs
```diff
@@ -59,6 +59,8 @@ pub struct FullTracks {
 /// relinking is applied.
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct SimplifiedTrack {
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub album: Option<SimplifiedAlbum>,
```
ramsayleung, line 63: So does it mean, when this field is none, Spotify just doesn't return this field rather than return this field with a null value?

channingbabb: > So does it mean, when this field is none, Spotify just doesn't return this field rather than return this field with a null value?

Correct.  I'm referencing the album track endpoint which has no album within the track (I'm assuming it's because you're using the album to get the track):
https://developer.spotify.com/documentation/web-api/reference/get-an-albums-tracks

ramsayleung (approved):

rspotify-model/src/track.rs
ramsayleung: Get it, thanks for your contribution. Merged :)

PR #402, opened 2023-04-23
Add ureq-native-tls feature flag

jirutka:
## Description

Allow building with ureq client and native-tls (i.e. without rustls).

## Motivation and Context

I’m an Alpine Linux developer and I want to build [ncspot](https://github.com/hrkfdn/ncspot) with native TLS instead of rustls.

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested?

```sh
cargo test --no-default-features --features client-ureq,ureq-native-tls
```

## Is this change properly documented?

Please make sure you've properly documented the changes you're making.

Don't forget to add an entry to the CHANGELOG if necessary (new features, breaking changes, relevant internal improvements).

PR #471, opened 2024-03-31
Fix ureq-native-tls feature to actually use native-tls

jirutka:
## Description

Feature `ureq-native-tls` currently doesn't work because ureq requires the TLS
connector to be specified explicitly for native-tls:

> `native-tls` enables an adapter so you can pass a
> `native_tls::TlsConnector` instance to `AgentBuilder::tls_connector`.
> Due to the risk of diamond dependencies accidentally switching on an
> unwanted TLS implementation, `native-tls` is never picked up as a
> default or used by the crate level convenience calls (`ureq::get`
> etc) – it must be configured on the agent.
> -- https://github.com/algesten/ureq#features

When ncspot is built with rspotify with ureq-native-tls, it's unable to
create an HTTPS connection:

	[ncspot::spotify_api] [DEBUG] http error: Transport(Transport { kind: UnknownScheme, message: Some("cannot make HTTPS request because no TLS backend is configured"), url: Some(Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("api.spotify.com")), port: None, path: "/v1/me/", query: None, fragment: None }), source: None })

## Motivation and Context

rspotify currently doesn’t work when built with `ureq-native-tls`.

## Dependencies 

`tls-native` (optional)

## Type of change

## How has this been tested?

CI

Yes

* * *
Related to #402

Fixes hrkfdn/ncspot#1159

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,6 +68,7 @@ jobs:
         features:
           - rspotify/cli,rspotify/env-file,rspotify/client-ureq,rspotify/ureq-rustls-tls,rspotify-http/client-ureq,rspotify-http/ureq-rustls-tls
           - rspotify/cli,rspotify/env-file,rspotify/client-reqwest,rspotify/reqwest-rustls-tls,rspotify-http/client-reqwest,rspotify-http/reqwest-rustls-tls
+          - rspotify/cli,rspotify/env-file,rspotify/client-ureq,rspotify/ureq-native-tls,rspotify-http/client-ureq,rspotify-http/ureq-native-tls
     steps:
       - name: Checkout sources
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


ramsayleung:
> `native-tls` enables an adapter so you can pass a
> `native_tls::TlsConnector` instance to `AgentBuilder::tls_connector`.
> Due to the risk of diamond dependencies accidentally switching on an
> unwanted TLS implementation, `native-tls` is never picked up as a
> default or used by the crate level convenience calls (`ureq::get`
> etc) – it must be configured on the agent.
> -- https://github.com/algesten/ureq#features

If I understand correctly, if you want to use `ureq-native-tls` feature, you have to enable `ureq-native-tls` feature and also explictly add `native-tls` as dependency and set it up?

ramsayleung (approved):

jirutka:
> you have to enable `ureq-native-tls` feature and also explictly add `native-tls` as dependency and set it up?

You have to explicitly pass a `native_tls::TlsConnector` instance to `AgentBuilder::tls_connector`. `native_tls` is provided by the `native-tls` crate, so yeah, you have to add `native-tls` as a dependency.

ramsayleung:
Thanks for the contribution.

ramsayleung:
Yeah, I've release the `v0.13.1`

PR #476, opened 2024-04-03
security: Disable TLS1.0/1.1 for ureq-native-tls

Property404:
## Description

Explicity opt-out of old TLS versions when using the ureq-native-tls feature. Rust-native-tls enables these outdated TLS versions by default

## Motivation and Context

TLS 1.0 and 1.1 have been deprecated since 2021. Enabling TLS 1.0/1.1 has possibility (albeit a low one) of introducing a security vulnerability. All modern clients and servers have supported 1.2 for years, so this should be an issue

## Dependencies 

None

## Type of change

- [ *] Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

Locally only, through cargo-test and cargo-clippy

Added a short comment explaining reasoning

### DIFF
--- a/rspotify-http/src/ureq.rs
+++ b/rspotify-http/src/ureq.rs
@@ -63,7 +63,11 @@ impl Default for UreqClient {
 
         #[cfg(feature = "ureq-native-tls")]
         let agent = agent.tls_connector(std::sync::Arc::new(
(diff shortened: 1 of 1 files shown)


ramsayleung (approved):

ramsayleung:

Issue #459, opened 2024-02-24
json parse error when getting playlist with no image


aome510:
**Describe the bug**
Calling `get-playlist` API with playlist having no image will return a response with `images` field `null`. However, according to https://docs.rs/rspotify-model/0.12.0/rspotify_model/playlist/struct.FullPlaylist.html, `rspotify` specifies the field as `Vec<Image>`. This leads to the following error when parsing the json response:

```
json parse error: invalid type: null, expected a sequence at line 1 column 293: invalid type: null, expected a sequence at line 1 column 293
```

**To Reproduce**
1. Create a new playlist with Spotify, making sure that it has no image
2. Call `get-playlist` API using the new playlist's URI
3. Observe `images` field has a value of `null`

ramsayleung:
I was unable to reproduce this problem:

```rust
#[maybe_async::test(feature = "__sync", async(feature = "__async", tokio::test))]
async fn test_playlist_with_no_image() {
    let playlist_id = PlaylistId::from_id("7G9P6fNlHvQ8lkObO5MJ7O").unwrap();
    let playlist = creds_client()
        .await
        .playlist(playlist_id, None, None)
        .await
        .unwrap();
    println!("playlist: {:?}", playlist);
    assert_eq!(playlist.images.len(), 0)
}
```

The `images` field from the response is empty rather than has a value of `null`:

```rust
[...]

PR #480, opened 2024-04-29
fix: deserialize empty images from null

plustik:
## Description

Adjust the derived implementation of `Deserialize` for `SimplifiedPlaylist`. Instead of returning an error when deserializing the value `null` for the field `images`, this value is deserialized to an empty `Vec`.
This is necessary, because the Spotify-API is returning `null` instead of an empty array for this field in some instances.

## Motivation and Context

This fixes the problem described in https://github.com/ramsayleung/rspotify/issues/459. The Spotify-API is returning `null` instead of an empty array for the `images`-field in some instances.

## Dependencies 

None

## Type of change

## How has this been tested?

The change was tested by using the patched version of rspotify in [this](https://gitea.tecks.eu/plustik/plservice) project.

I don't think this needs to be documented. The change won't affect the behavior of rspotify when the Spotify-API returns `images: []`. When the Spotify-API returns `images: null` `current_user_playlists` wasn't usable prior to this change. In this case the PR implements the expected behavior.

### DIFF
--- a/rspotify-model/src/playlist.rs
+++ b/rspotify-model/src/playlist.rs
@@ -20,6 +20,14 @@ pub struct PlaylistTracksRef {
     pub total: u32,
 }
 
+fn deserialize_null_default<'de, D, T>(deserializer: D) -> Result<T, D::Error>
+where
+    T: Default + serde::Deserialize<'de>,
+    D: serde::Deserializer<'de>,
+{
+    Ok(Option::deserialize(deserializer)?.unwrap_or_default())
+}
+
 /// Simplified playlist object
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


FichteFoll:
Would just using `#[serde(default)]` also work? https://serde.rs/attr-default.html

plustik:
> Would just using `#[serde(default)]` also work? https://serde.rs/attr-default.html

`#[serde(default)]` would not work. This is also discussed [here](https://github.com/serde-rs/serde/issues/1098). The workaround given there is almost the same as the one I copied from [here](https://github.com/ramsayleung/rspotify/issues/459#issuecomment-2025310972).

ramsayleung (approved):

ramsayleung:
Thanks for your contribution.

Issue #496, opened 2024-09-27
Search for multiple types


jakin010:
**Is your feature request related to a problem? Please describe.**
The spotify api allows for multiple type for search which is not available in the crate

**Describe the solution you'd like**
To allow multiple SearchTypes

PR #497, opened 2024-09-27
Added support for multiple search and added some tests

jakin010:
## Description

Their was no support for searching multiple types

## Motivation and Context

#496 

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

## How has this been tested?

- [ z] test_search_album: testing normal search
- [z ] test_search_multiple_types: testing if multiple types is returned

Yes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,10 @@
 
 **Bugfixes**
 - ([#494](https://github.com/ramsayleung/rspotify/pull/494)) Fix endless sequential pagination problem.
-
-## 0.13.4 (unreleased)
+ 
 **New features**
 - ([#496](https://github.com/ramsayleung/rspotify/pull/497)) Add support for searching multiple types
+
 
 ## 0.13.3 (2024.08.24)
 **New features**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 **Bugfixes**
 - ([#494](https://github.com/ramsayleung/rspotify/pull/494)) Fix endless sequential pagination problem.
 
+## 0.13.4 (unreleased)
+**New features**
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 **Bugfixes**
 - ([#494](https://github.com/ramsayleung/rspotify/pull/494)) Fix endless sequential pagination problem.
 
+## 0.13.4 (unreleased)
```
ramsayleung, line 8: Would you mind moving this entry to the 0.14.0?

src/clients/base.rs
```diff
@@ -454,6 +454,54 @@ where
...
+    /// - market - An ISO 3166-1 alpha-2 country code or the string from_token.
+    /// - include_external: Optional.Possible values: audio. If
+    ///   include_external=audio is specified the response will include any
+    ///   relevant audio content that is hosted externally.
+    ///
+    /// [Reference](https://developer.spotify.com/documentation/web-api/reference/#/operations/search)
+    async fn search_multiple(
+        &self,
+        q: &str,
+        _type: Vec<SearchType>,
```
ramsayleung, line 483: you could use `r#type` rather than `_type` to avoid keyword conflict.

ramsayleung: You don't need to constrain the collection type to `Vec`, it could be any type impl IntoIterator like this `impl IntoIterator<Item = SearchType>`, then you could pass a vec, a slice or other collection type conforming to `IntoIterator` trait.

ramsayleung (commented):
Thanks for your contribute, I left a few nitpick comments.

src/clients/base.rs
jakin010: Just made the changes

ramsayleung (approved):

PR #512, opened 2024-12-29
auth_code_pkce: add from_token_with_config() constructor

### LINKED ISSUE #511: No way to create AuthCodePkceSpotify from token with config
There is currently no way to create `AuthCodePkceSpotify` with a custom config.

Is there any reason to not have `AuthCodePkceSpotify::from_token_with_config()` as there is for `AuthCodeSpotify`?

gdesmott:
## Description

Add `AuthCodePkceSpotify::from_token_with_config()`

## Motivation and Context

This is required to be able to refresh an expired token from the cache as we need the client ID to refresh the token.

Fix #511

## Dependencies 

None

## Type of change

## How has this been tested?

- Call a `AuthCodePkceSpotify` using `with_config()` and

```rust
Config {
        token_cached: true,
        token_refreshing: true,
        ..Default::default()
    };
```

- Authenticate using the normal flow.
- Terminate the app and wait for one hour for the token to expired.
- Restore the token using `Token::from_cache()` and check that `token.is_expired()` returns `true`.
- Create a client with `AuthCodePkceSpotify::from_token()` and then call  `client.refresh_token()`. The client won't work as the token has not been refresh. You can also check by calling `client.refetch_token()` as it returns `None`.
- Now create a client using `AuthCodePkceSpotify::from_token_with_config()` passing the proper  `Credentials`. `client.refresh_token()` and `client.refetch_token()` are now working.

Yes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
  
 **New features**
 - ([#496](https://github.com/ramsayleung/rspotify/pull/497)) Add support for searching multiple types
+- ([#512](https://github.com/ramsayleung/rspotify/pull/512)) Add `AuthCodePkceSpotify::from_token_with_config()`
 
 
 ## 0.13.3 (2024.08.24)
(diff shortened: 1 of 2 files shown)


ramsayleung:
thanks for your contribution :)

ramsayleung (approved):

PR #526, opened 2025-05-26
Fix: deserialization of pagination with null items

DimitriTimoz:
## Description

Fix deserialization panic on the search endpoint when the API returns a list of elements containing one or more null values. Also fix CI (clippy warnings).

## Motivation and Context
When the Spotify API returns results for a search query, sometimes items are null, which causes a panic during deserialization.
Here is an example of a case that causes a panic:
```json
{
    "playlists": {
        "href": "https://api.spotify.com/v1/search?offset=17&limit=2&query=My%20Playlist&type=playlist&market=FR",
        "limit": 2,
        "next": "https://api.spotify.com/v1/search?offset=19&limit=2&query=My%20Playlist&type=playlist&market=FR",
        "offset": 17,
        "previous": "https://api.spotify.com/v1/search?offset=15&limit=2&query=My%20Playlist&type=playlist&market=FR",
        "total": 901,
        "items": [
            {
                ...
            },
            null // <--
        ]
    }
}
```

## Type of change

## How has this been tested?

I performed a search query on the FR market for playlists with the query "My Playlist". The response contained a null item. After applying the fix, I checked the same query again and confirmed that the panic no longer occurs (the "null" is still present in the response, but does not cause any error).

### DIFF
--- a/rspotify-model/src/page.rs
+++ b/rspotify-model/src/page.rs
@@ -22,6 +22,8 @@ pub struct Page<T: DeserializeOwned> {
     pub next: Option<String>,
     pub offset: u32,
     pub previous: Option<String>,
+    /// This field could mismatch the actual number of items in `items` field 
+    /// because sometimes the API returns `null` items that are not included in the `items` field.
     pub total: u32,
 }
 
--- a/rspotify-model/src/page.rs
+++ b/rspotify-model/src/page.rs
@@ -22,7 +22,7 @@ pub struct Page<T: DeserializeOwned> {
     pub next: Option<String>,
     pub offset: u32,
     pub previous: Option<String>,
-    /// This field could mismatch the actual number of items in `items` field 
+    /// This field could mismatch the actual number of items in `items` field
     /// because sometimes the API returns `null` items that are not included in the `items` field.
     pub total: u32,
 }
--- a/rspotify-model/src/page.rs
+++ b/rspotify-model/src/page.rs
@@ -1,11 +1,20 @@
 //! All kinds of page object
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
-use serde::{Deserialize, Serialize};
+fn vec_without_nulls<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
+where
+    T: serde::Deserialize<'de>,
+    D: serde::Deserializer<'de>,
+{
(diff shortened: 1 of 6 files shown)


rspotify-model/src/page.rs
```diff
@@ -1,11 +1,20 @@
 //! All kinds of page object
+use serde::{de::DeserializeOwned, Deserialize, Serialize};
 
-use serde::{Deserialize, Serialize};
+fn vec_without_nulls<'de, D, T>(deserializer: D) -> Result<Vec<T>, D::Error>
```
ramsayleung, line 4: Would you add a unit test for this function?

ramsayleung: And could you also add a comment for this function to demonstrate why do we need this custom deserialization function?

DimitriTimoz: I’m working on it, and I’m wondering how we handle the mismatch between the total number of results and the length of the `Vec`. `Null` elements are counted in the `total` field, but they are not present in the vector.

ramsayleung: We could accept this and add a doc to the `search` endpoint to demonstrate this anomaly behaviour. In my perspective, the Spotify server shouldn't return a `null` item, it doesn't make sense, and it requires all the clients and library to handle this weird design. What could the client do to handle `null` item except for filter them out.

tests/test_models.rs
```diff
@@ -1223,3 +1223,26 @@ fn test_null_id_in_tracklink() {
...
+  ],
+  "limit": 20,
+  "next": null,
+  "offset": 0,
+  "previous": null,
+  "total": 1
+}"#;
+    let page: Page<SimplifiedPlaylist> = deserialize(json);
+    assert_eq!(page.items.len(), 0);
+    assert_eq!(page.total-1, 0); // Because one element is null so ignore at deserialization but counted in total
```
ramsayleung, line 1244: We could just assert the `page.total` equal `1` to indicate we don't change the `total` field at all.

ramsayleung (approved):

DimitriTimoz:
@ramsayleung  the CI should pass know, I should have run the commands locally before.

ramsayleung:
Unfortunately, a job still failed, you could quickly fix it by:

```sh
cargo fmt -- --emit files
```

ramsayleung:

PR #574, opened 2026-08-10
bug fix: preserve refresh_token when auto-refreshing access token

blep:
## Motivation and Context

Spotify refresh token used for authentication become invalid after a few hours., causing all operations to fail.

## Description

Spotify's refresh-grant response may omit `refresh_token`; see
https://developer.spotify.com/documentation/web-api/tutorials/refreshing-tokens
"When a refresh token is not returned, continue using the
existing token." 

The bugfix preserve the previous refresh token so the next refresh
keeps working; otherwise the token is dropped on refresh and
the following refresh fails with no refresh token available.

This issue applies to both refresh_token() and prompt_for_token().

Notes: bug fix is AI generated (kilocode + DeepSeek V4 Flash High). I've only been using rust for a few months. I've fact checked the pointed out doc section, and code changes looks good to me.

## Type of change

- [X ] Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

This bug fix has been running in my app for 12h without issue or need to re-authenticate, whereas without it I needed to re-authenticate  every 1h-3h. 

Referenced relevant spotify doc in code, quoting the critical part.


src/clients/base.rs
```diff
@@ -85,7 +85,25 @@ where
     /// token will be saved internally.
     async fn refresh_token(&self) -> ClientResult<()> {
         let token = self.refetch_token().await?;
-        *self.get_token().lock().await.unwrap() = token;
+        let token_slot = self.get_token();
+        let mut token_lock = token_slot.lock().await.unwrap();
+        let merged = match (token, token_lock.as_ref()) {
```
ramsayleung, line 90: This CR looks good to me, but I have to say the API design is so weird from an API ergonomics perspective:

> Depending on the grant used to get the initial refresh token, a refresh token might not be included in each response. When a refresh token is not returned, continue using the existing token.

Why would the API caller need to distinguish the nuance and handle the absense and presense case of `refresh_token`, if the Spotify server want to indicate caller should keep using the existing token, the server could just return the existing token rather than omiting the filed.

ramsayleung (approved):

ramsayleung:
The CI jobs are failed

PR #576, opened 2026-09-16
fix: add default for FullTrack::external_ids

SheetMetalConnect:
### Problem

`GET /v1/me/player` and `GET /v1/me/player/currently-playing` return a track object without `external_ids`. `FullTrack::external_ids` has no default, so deserialization fails and the item falls back to `PlayableItem::Unknown`. Callers then lose the name, artists and album of the currently playing track.

Observed against the live API today, same account and token:

| Endpoint | `external_ids` present |
|---|---|
| `/v1/me/player` | no |
| `/v1/me/player?additional_types=track,episode` | no |
| `/v1/me/player?market=NL` | no |
| `/v1/me/player/currently-playing` | no |
| `/v1/tracks/{id}` | yes |
| `/v1/albums/{id}` | yes |
| `/v1/me/player/recently-played` | yes |

Every other required field of `FullTrack` is present in that payload, so `external_ids` is the only blocker. `popularity` and `available_markets` already got defaults in 0.16.0/0.16.1.

Downstream effect, for reference: spotify-player logs `WARN spotify_player::ui::playback: Unknown playback item` on every refresh and renders an empty playback line.

### Change

`#[serde(default)]` on `FullTrack::external_ids`, following the same approach as `FullArtist::genres` in #564. Not deprecated, since the field is still returned by the other endpoints.

Adds a regression test that deserializes a `CurrentPlaybackContext` whose track has no `external_ids`, and a CHANGELOG entry.


SheetMetalConnect:
Closing this: I submitted it without building or running the test. Reopening once it is verified against a build.

PR #578, opened 2026-09-16
fix: add default for FullTrack::external_ids

SheetMetalConnect:
## Description

`FullTrack::external_ids` has no default, but Spotify no longer sends the field on the track object returned by `GET /me/player` and `GET /me/player/currently-playing`. Deserializing a currently playing track therefore fails, and the item falls back to `PlayableItem::Unknown` — callers lose the name, artists and album of whatever is playing.

## Motivation and Context

Measured against the live API today, one account, one token:

| Endpoint | `external_ids` |
|---|---|
| `/v1/me/player` | absent |
| `/v1/me/player?additional_types=track,episode` | absent |
| `/v1/me/player?market=NL` | absent |
| `/v1/me/player/currently-playing` | absent |
| `/v1/tracks/{id}` | present |
| `/v1/albums/{id}` | present |
| `/v1/me/player/recently-played` | present |

So the field is not removed, only absent on the currently-playing endpoints, which is why this adds a default rather than a deprecation. Every other required field of `FullTrack` is present in that payload.

spotify-player carries a local workaround for exactly this — `patch_missing_track_external_ids` in `client/mod.rs` injects an empty `external_ids` into the JSON before handing it to rspotify. Fixing it here lets that shim go.

## Dependencies

CI on master currently fails on an unrelated clippy lint; #577 fixes that. This branch includes no fix for it, so its own CI run will stay red until #577 lands.

## Type of change

## How has this been tested?

- [x] `test_full_track_without_external_ids` added to `tests/test_models.rs`, using a real `/me/player` track payload
- [x] Full CI green on a fork with this commit plus the #577 lint fix: https://github.com/SheetMetalConnect/rspotify/actions/runs/35087474926

CHANGELOG entry added under `unreleased`.


ramsayleung (approved):
The CR looks good to me, but the CI jobs failed

ramsayleung (approved):
The CR looks good to me, but the CI job failed

